PR #1, opened 2018-10-25
Some suggestions

Skrypt:
...

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,3 +27,7 @@ Globalization should be taken into account at every step.
 Next steps will include:
 
 * Base shipping infrastructure, plus one implementation TBD
+* Product catalogs (prices affected by x factor: region, date)
+* Shopping cart rules (prices affected in cart by x factor: quantity, region, user role)
+* Shopping cart coupons (rebates applied globally)
+* Multicurrency support with real-time exchange rate conversion


Skrypt:
That would be a shopping cart discount rule based on user role.

bleroy:
Any particular reason not to do what Nwazet was doing for discounts, which is to have an [`IPromotion`](https://github.com/bleroy/Nwazet.Commerce/blob/master/Services/IPromotion.cs) interface? I understand there's room for improvement in the way this is applied, but the general idea should be the same, no?

bleroy:
Also, what's a product catalog? Is it doing anything commerce-specific, or could we instead re-use one of the Orchard constructs, such as taxonomies, to group products?

Having exchange rates integrated sounds like something we could have an extensibility point for, with maybe a static exchange rate implementation built-in.

Skrypt:
A product catalog is a set of products that can be sold on a specific tenant/store.

bleroy:
Yes, I'd rather not commit on that: I'm not sure we'll need it as a commerce-specific feature. The multi-store scenario is interesting in itself though, and it seems like one more of those where data sharing between tenants could be useful. Not sure what the thinking is for that sort of thing in Core at this day.

bleroy:
I like those. How about we create a roadmap wiki page to capture those? Something like [this](https://github.com/OrchardCMS/OrchardCore.Commerce/wiki)?

MatteoPiovanelli-Laser:
This post will likely go off topic. Feel free to take it apart and move different sections to their own thread/wiki page, or tell me where I should move things.

Some ideas on discounts/coupons/promotions from notes I have about stuff I haven't gotten around to implementing for Nwazet. A lot of this will be obvious. I'll mostly talk about promotions, meaning also coupons and discounts.

 - Promotion can be on a single product, on a cart "line", or on the whole cart.
 - Promotion has a condition that says whether it applies at the time. Some examples: user is authenticated, user belongs to role, user has entered a code, it is a given date, some combination/quantity of products is in the cart, total price on the cart is above a threshold...
 - Promotions can also be applied on or have to do with shipping. e.g.: free shipping for orders above X.
 - Coupons can expire, be private (to a single user or group).
 - Conditions related to prices may be based on the price before tax, after tax, before shipping...
 - Promotion may have conditions that relate them to each other. E.g.: a promotion that may be applied only if no other promotion is applied.
 - Promotions have a way to sort them, like a priority. I think one should always first do product or line promotions, then cart promotions, but in either group there may be other factors involved. E.g.: how to handle the case where a coupon decreases the price for a single unit of a product, dropping the subtotal under the threshold above which we would apply a cart discount, or free shipping?

An additional note about taxes:
Tax can make a mess of things: in the EU, the user must be shown from the very beginning the price including tax. I think in the US the user looks at the price before tax. That may affect the values on which promotion conditions should apply.
In talks with a local (here in Italy, but regarding EU taxes) accountant/lawyer, they said that basically the only company that is handling things correctly internationally is Dell. Other companies (he made the example of Amazon explicitly at the time) are basically evading a bunch of taxes "on purpose" to apparently simplify things for customers, and paying fines afterwards.

I think all the features mentioned by @jeffolmstead should be in, or at least planned for, as well as the multi-store thing.
The "e-commerce" examples I have in mind and that the module should enable "painlessly" are:
 - amazon.com: digital and physical products. Subscription.
 - drivethrurpg.com: both digital and physical products. Pay-what-you-want, free, fixed price products. Single products with different purchase options: physical copy, digital only, digital+physical... Product "updates". Download again stuff bought in the past. Additional checkout steps: before download, each digital product is watermaked as yours. Not exactly a multi-store, but "anyone" can put a product there.
 - ko-fi.com / patreon.com: fixed/configurable donations. Multi-store. Subscription. You are really buying nothing, or access to content.
 - booking.com: the definition of products here includes a time period. When I am booking a room in a hotel from Friday to Tuesday, what is really the product?
 - ticketone.it: tickets for shows and stuff. Product definitions include the time, but unlike the booking case, it is a fixed and specific time, rather than a period that can be configured.
 - ebay.com: multi-store. Auctions.
I don't expect any of them to start using Orchard, but their features should be possible. I think we should keep those in mind where designing features so we don't prevent scenarios, or at least know when we do.

bleroy:
Thanks for all the excellent ideas... that we should probably not discuss in a pull request... :)

I'm going to suggest the following workflow: feature ideas and suggestions are created as issues, so we can debate and comment them, then once we have some sort of consensus, we move them to the wiki as semi-formal designs. Then, or at the same time, implementation.

What do you think? Can I close this PR for now, and we begin to move all those ideas to issues? Im also going to remove speculation about non-MVP features from the readme, into issues as well.

PR #26, opened 2019-04-16
PricePart default currency

microposmp:
I've implemented a few small changes for the PricePart.

- Fixed a NullReferenceException when creating a new content item that contains a PricePart.
- Added SEK (Swedish krona) currency. Thinking about making an editor for maintaining currency data.
- Added a settings page with option to configure default currency.

I'm interested in participating in the development of the OrchardCore.Commerce module.

### DIFF
--- a/Services/CommerceSettingsConfiguration.cs
+++ b/Services/CommerceSettingsConfiguration.cs
@@ -1,0 +1,33 @@
+using System;
+using Microsoft.AspNetCore.DataProtection;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using OrchardCore.Commerce.Abstractions;
+using OrchardCore.Entities;
+using OrchardCore.Settings;
+
+namespace OrchardCore.Commerce.Services
+{
+    public class CommerceSettingsConfiguration : IConfigureOptions<CommerceSettings>
+    {
+        private readonly ISiteService _site;
+        private readonly ILogger<CommerceSettingsConfiguration> _logger;
+
+        public CommerceSettingsConfiguration(
+            ISiteService site,
+            ILogger<CommerceSettingsConfiguration> logger)
+        {
+            _site = site;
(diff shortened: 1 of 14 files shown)


Services/CommerceSettingsConfiguration.cs
```diff
@@ -0,0 +1,36 @@
...
+using OrchardCore.Commerce.Abstractions;
+using OrchardCore.Entities;
+using OrchardCore.Settings;
+
+namespace OrchardCore.Commerce.Services
+{
+    public class CommerceSettingsConfiguration : IConfigureOptions<CommerceSettings>
+    {
+        private readonly ISiteService _site;
+        //private readonly IDataProtectionProvider _dataProtectionProvider;
```
bleroy: Please delete, rather than comment out.

bleroy:
If you could also squash out those gitignore changes, to keep history clean, that would be great. Thanks!

bleroy:
Great thanks. Unless you're going to do it yourself, I'll squash some of those before I merge.

microposmp:
Hi, I don't know how to do the squash. I'll learn that for future commits. Can you do it for this change? Thanks

bleroy:
Sure, no problem. I'll do it, that's perfectly fine. FWIW, the way you do it is that you do an interactive rebase, something like `git rebase -i HEAD~7` where 7 should be replaced with how many changes you want to act on. Then you get a text editor with the list of changes, and on each line you can replace the operation according to the comment that got inserted in there, to squash changes with the ones above (I used 'f' the most, personally). Once you're satisfied with the result, and you have a clean change list, you do a push of the branch with the `--force-with-lease` option. That last one is what rewrites history on the github side.

Services/CommerceSettingsConfiguration.cs
microposmp: Removed commented out code and squashed the commit into the previous commit.

PR #27, opened 2019-04-17
Product recipe

agriffard:
I tested with an OC instance setup from a Saas recipe which has only a few modules enabled.

The feature needs a dependency on OrchardCore.Contents in the Manifest or else, enabling it will produce an error.

Then, I defined a `Product` content type with those parts:
Title
Autoroute
Html
Product
Price

and created 2 products.

Finally, I exported it as a recipe and made it available with a Product.recipe.json file that will enable the required features and create the content type and the sample data.


agriffard:
Ok, samples removed.

I was also wondering whether adding a 'Product Image' Media picker.
In that case, a Product detail and summary templates (or specific to the ProductImage) could be interesting.

bleroy:
Maybe, it seems pretty unlikely that you would *not* want an image, and for those cases, it's easy to remove. Eventually, I'd also like to have a separate sample recipe that sets up a complete sample site, with a catalog, etc. That one is different: it's not a starting point, it's a sample. For your recipe, we should focus on most common requirements. Cheers!

Issue #25, opened 2019-04-06
Product attributes


bleroy:
Product attributes (such as sizes, colors, dimensions, and other options) need to be baked early into the system, as their presence affects the signature of important APIs, not least among them shopping cart APIs.

Product attributes are variants of a product that the customer can choose when adding a product to the cart. They are of arbitrary types (one can imagine enums, measurements, or even text), and can affect pricing.

They share a lot of characteristics with Orchard fields, in that they are named, can be added an arbitrary number of times to a specific product type, and can be configured. The main difference is that the actual value is not attached to the product but to the shopping cart item. There is enough in common that attributes in Core Commerce will be built as new Orchard fields. The actual attribute values will be separate objects that travel alongside the product reference in the shopping cart. The field is what is used to *configure* the attribute.

Of course, this is extensible, and anybody can implement their own attribute field. One could imagine building whole configurators for example.

bleroy:
As a first step, I'm implementing the following attribute field types:

* Boolean
* Numeric
* Text
* Enumeration

bleroy:
Q: should the text/enum attribute allow for per product allowed values overrides?

bleroy:
I am not going to include price adjustments to the implementation of these simple attributes. There will be two ways of implementing them: you could build your own attributes that do include price adjustments (thus imposing your own rules), or you could build price providers that could map attribute values to price adjustments.

The guiding principle here is to keep extensibility points for the more complex scenarios, but to keep the default implementation as simple as possible while enough for simple scenarios.

bleroy:
I am not going to include price adjustments to the implementation of these simple attributes. There will be two ways of implementing them: you could build your own attributes that do include price adjustments (thus imposing your own rules), or you could build price providers that could map attribute values to price adjustments.

The guiding principle here is to keep extensibility points for the more complex scenarios, but to keep the default implementation as simple as possible while enough for simple scenarios.

bleroy:
#31 has a discussion that is relevant to this.

PR #30, opened 2019-04-22
Introduce product attributes

bleroy:
Product attributes are defined as fields that usually have nothing to edit in the product editor. The parameters of the attributes are usually set-up at the content type definition level, as settings of the fields.

Built-in product attributes are Boolean, numeric, and text (which includes predefined values, and thus can handle enums, flags, and combos). This should be enough for most cases, but is, as usual, extensible.

The front-end for those new fields will be introduced with the shopping cart: each attribute will participate as adding information to the "add to cart" action. For example, adding two shirts of size M.

### DIFF
--- a/Drivers/ProductAttributeFieldDriver.cs
+++ b/Drivers/ProductAttributeFieldDriver.cs
@@ -1,0 +1,60 @@
+using Microsoft.Extensions.Localization;
+using OrchardCore.Commerce.Fields;
+using OrchardCore.Commerce.Settings;
+using OrchardCore.Commerce.ViewModels;
+using OrchardCore.ContentManagement.Display.ContentDisplay;
+using OrchardCore.ContentManagement.Display.Models;
+using OrchardCore.DisplayManagement.Views;
+
+namespace OrchardCore.Commerce.Drivers
+{
+    public abstract class ProductAttributeFieldDriver<TField, TFieldSettings> : ContentFieldDisplayDriver<TField>
+        where TField : ProductAttributeField, new()
+        where TFieldSettings : ProductAttributeFieldSettings
+    {
+        public ProductAttributeFieldDriver(
+            IStringLocalizer<ProductAttributeFieldDriver<TField, TFieldSettings>> localizer)
+        {
+            T = localizer;
+        }
+
+        public IStringLocalizer T { get; set; }
+
+        public override IDisplayResult Edit(TField field, BuildFieldEditorContext context)
+        {
+            return Initialize<EditProductAttributeFieldViewModel<TField, TFieldSettings>>(
+                GetEditorShapeType(context), model =>
+                {
+                    var settings = context.PartFieldDefinition.Settings.ToObject<TFieldSettings>();
+                    model.Field = field;
+                    model.Settings = settings;
+                    model.Part = context.ContentPart;
+                    model.PartFieldDefinition = context.PartFieldDefinition;
+                });
+        }
+    }
+
+    public class BooleanProductAttributeFieldDriver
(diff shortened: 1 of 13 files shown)


agriffard:
The editor of the settings of the TextProductAttributeField does not seem to be displayed when I try to configure it.

bleroy:
Works fine for me. Anything in logs?

agriffard:
Yes, I should have seen that:

agriffard:
Yes, I should have checked this:

```
2019-04-22 13:41:59.0368|Default|8000004b-0004-fe00-b63f-84710c7967bb||OrchardCore.ContentTypes.Editors.IContentDefinitionDisplayHandler|ERROR|IContentPartFieldDefinitionDisplayDriver thrown from OrchardCore.Commerce.Settings.TextProductAttributeFieldSettingsDriver by ArgumentNullException System.ArgumentNullException: Value cannot be null.
Parameter name: value
   at System.String.Join(String separator, String[] value)
   at OrchardCore.Commerce.Settings.TextProductAttributeFieldSettingsDriver.<>c__DisplayClass0_0.<Edit>b__0(TextProductAttributeSettingsViewModel viewModel) in D:\Orchard\OrchardCoreDev\src\OrchardCore.Modules\OrchardCore.Commerce\Settings\ProductAttributeFieldSettingsDriver.cs:line 50
```

bleroy:
Good catch, I'll fix it now. Thanks.

bleroy:
Done. Please check again. Thanks!

agriffard:
Ok, it works now. Thank you.

@Skrypt managed to use a nice editor for predefined values on a text field.

bleroy:
Sure, I'm going for MVP for the moment, but I'll gladly merge PRs making stuff nicer.

microposmp:
The settings view for BooleanProductAttributeField is missing a field for configuring default value.

image

Added below code block to test the logic.

```html
<fieldset class="form-group">
    <div class="row col-sm">
        <label asp-for="Label">@T["Default value"]</label>
    </div>
    <div class="custom-control custom-checkbox">
        <input asp-for="DefaultValue" type="checkbox" class="custom-control-input">
        <label class="custom-control-label" asp-for="DefaultValue">@T["On/Off"]</label>
    </div>
    <div class="row col-md">
        <span class="hint">@T["The default value associated with this attribute."]</span>
    </div>
</fieldset>
```

bleroy:
Good catch, thanks!

Skrypt:
Just saying but I think you can also use Bootstrap switch toggles.
image

microposmp:
Now I think it's clear to me what the real purpose of the ProductAttributeFields are. I missed this sentence in the description in #25. "_The main difference is that the actual value is not attached to the product but to the shopping cart item._". 

I really like this feature. It enables easy configuration of scenarios that we today call additional handling.  We today see multiple scenarios where this would be useful. For example printing custom information on items.

In the sample above with the Discountable field I would use the standard BooleanField in OrchardCore. I like the idea with the default value configurable and see value in making it part also in the OrchardCore standard BooleanField. I'll do that as an exercise.

I don't see the need to consider the ProductPart (SKU) as I described in #31 for the ProductAttributeFields. I think size/color on products in stock (inventory) should be handled in a different way. I'll try to clarify how I think about this with some samples. Plan to do it in the weekend.

bleroy:
Thanks. I do think the possibility for different SKUs need to be there at least as an extensibility point: one of the biggest challenges in designing an e-commerce platform, as opposed to a one-shot e-commerce web site, is that no two businesses share exactly identical processes.

The way I see it, the SKU could be modified by any service implementing a certain interfaces. Attributes would be one way to affect it, but there could be others.

PR #34, opened 2019-05-12
Currency as content items

microposmp:
Added a new CurrencyProvider that handles the currency as content items. This enables the addition of currencies without recompiling.

Added a Admin menu stub with the layout shown below. A recipe is provided that sets up the menu and currency list. US Dollar is created as currency in the recipe.

- **General** General commerce module settings. Currently only the default currency option is available there.
- **Currencies** List where maintenance of currencies can be performed.

image

There are still some more things to do before it's complete:
- [x] Permissions need to be comfigured.
- [x] A restart is required to make newly added currencies visible/selectable.

Thinking about similar configurations for tax rule, size curves etc.

Later on a different theme would be nice.

Is this in the right direction or would it be better to store this kind of data in a different way?


Indexes/CurrencyPartIndex.cs
```diff
@@ -0,0 +1,50 @@
...
+using OrchardCore.Commerce.Money;
+using OrchardCore.ContentManagement;
+using YesSql.Indexes;
+
+namespace OrchardCore.Commerce.Indexes
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public class CurrencyPartIndex : MapIndex
```
bleroy: Why do we need an index on currencies? That seems wrong to me, especially as there may be multiple `ICurrencyProvider` implementations all contributing currencies.

Migrations/CurrencyMigrations.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        public CurrencyMigrations(IContentDefinitionManager contentDefinitionManager)
+        {
+            _contentDefinitionManager = contentDefinitionManager;
+        }
+
+        public int Create()
+        {
+            _contentDefinitionManager.AlterTypeDefinition("Currency", builder => builder
+                .WithPart("TitlePart")
```
bleroy: Why do we need the title part?

bleroy:
I'd also prefer to see dynamically created currencies as some sort of last resort fallback. Currencies are after all such unchanging entities (not talking about exchange rates, which are out of scope for this discussion). I'd prefer a specialized admin screen that also shows currencies from non-dynamic providers as read-only, but does show their presence. Finally, why keep the default currency setting in general settings if we're going to have a currencies settings screen?

bleroy:
The idea with currency providers was that the core module could provide the most common ones (like it does), and have some optional currency pack(s) providing the rest as needed, and a dynamic provider such as yours for the rare case where a new one must be added.

Migrations/CurrencyMigrations.cs
microposmp: I agree that this is not needed. It was added for better display in the list view. 

If I override the summary view for the currency type the Name property could be shown instead.

bleroy:
I'd also isolate the dynamic currency provider in a separate feature.

microposmp:
I'll do adjustments according the feedback above.

I saw value in having it dynamic so it's easy to start in all regions. Maybe it's not as valuable because it will in most situations require coding and then it's easy to add the value in code. I'll complete it if it's interesting to bring it into the product. I'll probably complete it anyway as it's a small component that improves my experience with OrchardCore CMS.

- [ ] Add to a separate feature.
- [ ] Storage / Invalidation
- [ ] Remove TitlePart and update admin summary view for currency.

Thanks for the feedback. I'll continue tomorrow. It's almost midnight where I live.

Are there any areas where I can provide help? I need to start with some easy tasks to learn more about OrchardCore development.

bleroy:
Thanks for the contribution and for the offer to do more. I should at some point add a list of possible contributions. It's tricky at this point because we're still very early. I'd say maybe look at the stuff that is done (prices, money, like you've done, attributes), and see if you find additional features and extensions that would be useful to you or you'd like to add, then file a bug for discussion. Once the MVP is in place, there'll be a lot more open for contributions.

Migrations/CurrencyMigrations.cs
microposmp: Name is now stored as DisplayText on the content item. TitlePart is removed.

Indexes/CurrencyPartIndex.cs
microposmp: I added the index so that it would be quick to find the currency documents when it's stored as content items.

I ran into some issues when testing with loading multiple CurrencyProviders at the same time. I'll write more about that below in the main thread.

microposmp:
In the below description Currency is only an example. It can apply to any type of data and I think it's important to consider these cases early in the design.

I did some tests with configuring multiple currency providers and loading them. In scenarios like the following I see some issues with the current design.

1. There is the default provider with a base set of currencies defined.
2. A region/country package is built that defines new currencies for that region/country.
For example: Country package for Norway implements a currency provider that adds NOK as currency.
3. When integration a payment provider I also see that as a possible place to add currency providers.

The following combination for NOK might result in conflicts that have to be resolved. 

1. Default currency provider. NOK is not available here. 
2. Country package for Norway adds a local currency provider that adds NOK.
3. Payment provider integration that also adds a local currency provider that supports all Nordic countries (DKK (Denmark), NOK (Norway), SEK (Sweden) and EUR (Finland)).

Which provider should win and provide NOK in the above configuration?

In scenarios like the above I think it's better to have configuration options instead of hardcoded providers.

I'd also like the possibility to clear out all currencies except one for a small local web shop that only operates in one country.

I think one dynamic provider is the most flexible solution for the shop/site administrator.

With the multiple provider solution and static currency data configuration will also be needed for resolving possible conflicts.

The above applies even if the data that is configured is very slow changing.

**In other scenarios I see the multiple providers as a great idea.** 

For example:
- Discounts and Promotions. New rules can be added by installing new packages that provide calculation rules.
- Providers for tax rules.
- Support for multiple payment providers.
- Many more ...

bleroy:
I don't understand. What differences would there be between the two NOK currencies provided? Why does it matter which one wins?

bleroy:
I also still don't see what benefit there is in currencies being content items.

microposmp:
It don't have to be content items, but they are very flexible and easy to implement. It could be stored in another way. 

Currently there will be multiple entries in the dropdowns when a currency is defined in multiple providers. That can be solved easily by filtering in the MoneyService.

But there is also the possibility that 2 providers define the values for en entry differently. If the filter doesnät always return the same entry data can be shown randomly. Hopefully all components that define providers return the same entry identically.

Maybe I'm thinking to much about this, but I think it should only be one truth of the data.

bleroy:
Here's what I have in mind: currencies are *simple* structs that are bare representations of currency data. They are essentially universal singleton types of entities, like numbers or physical units. There should be only one correct way of defining each of them, so if there are duplicates, they should be identical, unless one of them is plain wrong.

For those reasons, I'd store the dynamically defined currencies as plain data stored in a site setting. I'd also filter them so they are unique by symbol (and it shouldn't matter which one is selected, they should be identical even if duplicated).

microposmp:
Ok, do you want me to do the following?

- Change the DefaultCurrencyProvider to be dynamic.
- Change storage to site settings.
- Make a view for maintaining.
- Change to only have one provider active in MoneyService.

That way it should only be one possible configuration for each entry and it's dynamically configurable.

microposmp:
Or another idea:

Define all currencies currently available hardcoded in the default provider and store which should be enabled on the site in site settings.

The chainges I think will be needed for that is:

- Hardcode all currencies in DefaultCurrencyProvider
- Provide configuration screen where currencies can be enabled. 
- MoneyService will filter and show currencies that are enabled.

If a currency is changed it will require program update. But it's not very commonly changed.

I like the dynamic solution better. It's a bit more work, but will not require any changes of hardcoded values going forward.

bleroy:
Why do currencies need to be disabled?

microposmp:
My thought was to make entry of product data and so on as simple and error free as possible. If only the options that should be possible to select there will be fewer errors in the data going out to customers. If there are lots of possible entries to select there will be more errors.

For example if you only sell in Sweden (SEK) there is no need to select any other currency when entering a price.

Most larger sites will probably import all data or make lookups directly against a backend system and don't make much maintenance in the site admin, but many smaller sites might do all product maintenance in the product. 

The idea is to make it as simple as possible by not showing options that are not valid in the market the site is deployed.

There is no absolute requirement.

I'll stop for today now. Almost midnight again and time to sleep. I'll look at this again later in the week.

bleroy:
If you sell only in Sweden, wouldn't you just make SEK the default? G'night.

bleroy:
Hey! I hope I didn't discourage you.

Thinking a little more about this, I think another interesting analogy to consider is culture. As rarely changing entities, they are defined as part of the framework, but you can still create your own in the rare cases where it's necessary. In Orchard, you can select which ones you want to use.

So I think similarly, your idea to have UI to choose which currencies you want to use is a good one, but there's still no reason why they should be content items. I think the sweet spot here is to have them as simple structures like today, have the default provider define all of them, and still have the extensibility point for the rare cases. Then, have a setting that stores the list of active currencies (a simple list of currency symbols stored directly in settings will do).

microposmp:
No problem, there have been much to do at my daily work.

I'll put build an updated suggestion where all currencies will be defined in the default provider.
I'm also thinking about some configuration for the PricePart regarding currencies.

Hope to update the pull request with these changes during next week.

bleroy:
Any update on this, @microposmp ?

bleroy:
Bump cc @microposmp

microposmp:
I'll look into this again and see how it can work with the changes that have been made.

microposmp:
I think we should close this pull request. I'll start a new one with extended configuration settings for currencies when I have something to share that will build on the current implementation. Preferably the changes provided in #49.

What I think of is to enable configuration of what currencies should be enabled. Something like the culture settings in Orchard Core.

PR #39, opened 2019-10-28
Shopping cart

bleroy:
This is a complete implementation of an attribute-aware shopping cart.

Note that this is almost entirely devoid of javascript, by design. Later on, we'll have incremental cart improvements, such as a shopping-cart widget that can be on all pages and that supports Ajax updates, as well as alternative cart storage, including client storage.


Views/NumericProductAttributeFieldSettings.Edit.cshtml
```diff
@@ -35,31 +35,51 @@
 <fieldset class="form-group">
```
Skrypt: use div instead of fieldset when using class="form-group" like this.

agriffard:
Do you also have this kind of display in `SummaryAdmin` for a content with a ProductPart?

image

Views/ShoppingCart/Index.cshtml
```diff
@@ -0,0 +1,88 @@
...
+                <div class="intro-heading text-uppercase">@T["Your cart"]</div>
+            </div>
+        </div>
+    </header>
+</zone>
+<section id="shopping-cart">
+    @if (Model.Lines != null && Model.Lines.Any())
+    {
+    <table class="table">
+        <thead><tr><th class="col-1">@T["Quantity"]</th><th class="col-9">@T["Product"]</th><th class="col-1">@T["Price"]</th><th class="col-1">@T["Action"]</th></tr></thead>
```
agriffard: `col-` css classes do not exist in Bootstrap 4.3.1 for `<th>`.

In order to set the width of each column, you need to use inline style or additional css classes that will do the equivalent, like this basic sample:

```
<th style="width:10%">Quantity</th>
<th style="width:70%">Product</th>
<th style="width:10%">Price</th>
<th style="width:10%">Action</th>
```

jeffolmstead: Don't want to detract too much,  but is there any chance the checkout pave can move away from using a table architecture? I don't know what the alternative is, I just know in Orchard I had to make all the tables scroll left to right otherwise mobile users had a difficult time checking out.

Skrypt: Mini cart summary could work. Look at how Shopify does it.

bleroy:
> Do you also have this kind of display in `SummaryAdmin` for a content with a ProductPart?

Yes, that's my last minute addition of a placement file. I'll fix that.

Views/ShoppingCart/Index.cshtml
Skrypt: Good distinction to make 😉

bleroy: @jeffolmstead I think I'll give it a try tonight. <rant>It's slightly depressing that HTML went from unsemantic tables for everything to divs + CSS to semantic markup, and back to unsemantic divs with semantics defined in CSS.</rant> :D

Skrypt: You could try Flexboxes.

bleroy: I ripped tables out of the cart.

PR #40, opened 2019-10-28
Update product recipe

agriffard:
Use the latest OC recipe format.


agriffard:
I should not have created it from your branch.

PR #48, opened 2020-02-05
Currency serialization fixes

microposmp:
Fix reading known currencies and add back initialization of default currency configured in settings.

I've only done tests with the PricePart, The changes are not verified with the PriceVariantsPart.

The main issue was that when a known currency was saved only the CurrencyIsoCode is saved. When reading it back the missing values resulted in it being handled as unknown.

Added back the CommerceSettingsConfiguration so that the default currency configured for the site is taken into consideration.

### DIFF
--- a/Drivers/PricePartDisplayDriver.cs
+++ b/Drivers/PricePartDisplayDriver.cs
@@ -1,4 +1,5 @@
 using System.Threading.Tasks;
+using Money;
 using OrchardCore.Commerce.Abstractions;
 using OrchardCore.Commerce.Models;
 using OrchardCore.Commerce.ViewModels;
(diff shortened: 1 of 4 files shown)


Drivers/PricePartDisplayDriver.cs
```diff
@@ -48,7 +48,7 @@ private Task BuildViewModel(PricePartViewModel model, PricePart part)
 
             model.Price = part.Price;
             model.PriceValue = part.Price.Value;
-            model.PriceCurrency = part.Price.Currency?.CurrencyIsoCode ?? _moneyService.DefaultCurrency.CurrencyIsoCode;
+            model.PriceCurrency = part.Price.Currency?.CurrencyIsoCode == "---" ? _moneyService.DefaultCurrency.CurrencyIsoCode : part.Price.Currency?.CurrencyIsoCode;
```
bleroy: Can you compare with the unspecified currency's symbol explicitly instead of copying its magic string?

MoneyDataType/LegacyAmountConverter.cs
```diff
@@ -21,7 +21,7 @@ public override Amount ReadJson(Newtonsoft.Json.JsonReader reader, Type objectTy
             string symbol = null;
             string iso = null;
             int? dec = null;
-            bool unknown = false;
+            //bool unknown = false;
```
bleroy: Don't comment out code. If it must be deleted, delete it.

microposmp: Commented out lines are deleted.

Drivers/PricePartDisplayDriver.cs
microposmp: Comparison changed to be made against UnspecifiedCurrency.

PR #49, opened 2020-02-08
Native currency names

microposmp:
Changed to use the english currency name. We got random results with the name on different machines.

### DIFF
--- a/MoneyDataType/KnownCurrencyTable.cs
+++ b/MoneyDataType/KnownCurrencyTable.cs
@@ -40,7 +40,7 @@ namespace Money
                 CurrencyTable.Remove(currentCultureCurrency.CurrencyIsoCode);
                 CurrencyTable.Add(currentCultureCurrency.CurrencyIsoCode, currentCultureCurrency);
 
-                CurrencyTable.Add("BTC", new Currency("BitCoin", "₿", "BTC", 8));
+                CurrencyTable.Add("BTC", new Currency("BitCoin", "BitCoin", "₿", "BTC", 8));
             }
         }
 
--- a/MoneyDataType/KnownCurrencyTable.cs
+++ b/MoneyDataType/KnownCurrencyTable.cs
@@ -1,4 +1,4 @@
-﻿using System.Collections.Generic;
+using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
 using Money.Abstractions;
@@ -34,6 +34,11 @@ namespace Money
                     .Select(c => new Currency(c)).Cast<ICurrency>()
                     .Distinct(new CurrencyEqualityComparer())
                     .ToDictionary(k => k.CurrencyIsoCode, e => e);
+
+                // Make sure the CurrentCulture currency is used for countres with multiple cultures/languages.
+                var currentCultureCurrency = new Currency(CultureInfo.CurrentCulture) as ICurrency;
+                CurrencyTable.Remove(currentCultureCurrency.CurrencyIsoCode);
+                CurrencyTable.Add(currentCultureCurrency.CurrencyIsoCode, currentCultureCurrency);
 
                 CurrencyTable.Add("BTC", new Currency("BitCoin", "₿", "BTC", 8));
             }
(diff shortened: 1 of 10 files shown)


bleroy:
What kind of issue did you get? This looks wrong.

microposmp:
The issue is that the call to CultureInfo.GetCulture returns cultures in different order on different machines.

In the function InitCurrencyCodeTable this introduces a problem when distinct currencies are retrieved.

```csharp
private static void InitCurrencyCodeTable()
{
	lock (Obj)
	{
		bool valid(CultureInfo c) => !c.IsNeutralCulture && !c.EnglishName.StartsWith("Unknown Locale") && !c.EnglishName.StartsWith("Invariant Language");

		CurrencyTable = CultureInfo.GetCultures(CultureTypes.AllCultures)
			.Where(valid)
			.Select(c => new Currency(c)).Cast<ICurrency>()
			.Distinct(new CurrencyEqualityComparer())
			.ToDictionary(k => k.CurrencyIsoCode, e => e);

		CurrencyTable.Add("BTC", new Currency("BitCoin", "₿", "BTC", 8));
	}
}
```

The code block below returns only the cultures that have SEK as currency. On different machines (with the same user and CurrentCulture "sv-SE") this is returned in different order.

```csharp
var cultures = CultureInfo.GetCultures(CultureTypes.AllCultures)
	.Where(valid)
	.Select(c => new Currency(c))
	.Where(c => c.CurrencyIsoCode == "SEK")
	.ToDictionary(k => k.Culture, e => e);
```


image

The currency is the same, but the native translation is different based on the culture. Will provide a new commit where the CurrentCulture currency is always used regardless of retrieval order.

bleroy:
Still unclear why this is a problem. Furthermore, the proposed fix also looks arbitrary and unreliable.

microposmp:
The problem for me was that I couldn't find Swedish Krona and configure it as default currency. The first loaded culture has a native name that I don't understand for the currency SEK.

Made some more changes to make it more reliable and should now offer a good startup experience in most regions.

- Changed to use the system CurrentCulture currency as the DefaultCurrency if no default is configured. The first thing I do after enabling the commerce feature is go into settings and change to our local currency. I think it's a good default to use the system CurrentCulture currency as the default. It's still possible to override in commerce settings.
- Now reads both native and English currency names from system region settings. The English name can be translated via the OrchardCore localization feature.
- I've not found anywhere the native name is currently shown but kept the logic to replace the first loaded currency where the region has the same currency as the CurrentCulture currency. This is to make the currency name understandable to the user.

MoneyDataType/KnownCurrencyTable.cs
```diff
@@ -35,7 +35,12 @@ private static void InitCurrencyCodeTable()
                     .Distinct(new CurrencyEqualityComparer())
                     .ToDictionary(k => k.CurrencyIsoCode, e => e);
 
-                CurrencyTable.Add("BTC", new Currency("BitCoin", "₿", "BTC", 8));
+                // Make sure the CurrentCulture currency is used for countres with multiple cultures/languages.
+                var currentCultureCurrency = new Currency(CultureInfo.CurrentCulture) as ICurrency;
+                CurrencyTable.Remove(currentCultureCurrency.CurrencyIsoCode);
+                CurrencyTable.Add(currentCultureCurrency.CurrencyIsoCode, currentCultureCurrency);
```
bleroy: I still don't understand how this works. Why would you replace this in a static method? You can't know that what is the current culture when this runs (which is pretty much non-deterministic) is going to remain the current culture. The current culture potentially changes all the time. The whole idea of defaulting to the current culture sounds good on the surface, but I don't think it holds water. The default culture of the site may be a better choice, but even if you do that, you would certainly not modify the table (especially not in a static method: all tenants share that if I'm not mistaken), you'd instead select the correct current culture when it's queried for.

Services/MoneyService.cs
```diff
@@ -35,7 +36,7 @@ public ICurrency DefaultCurrency
             {
                 var defaultIsoCode = _options?.DefaultCurrency;
                 return string.IsNullOrEmpty(defaultIsoCode)
-                    ? Currency.USDollar
+                    ? new Currency(CultureInfo.CurrentCulture)
```
bleroy: See my other comment.

MoneyDataType/KnownCurrencyTable.cs
microposmp: NativeName is currently never shown so this change is rolled back.

Services/MoneyService.cs
microposmp: I agree with this. Have rolled back this change.

bleroy:
Thanks for the contribution!

PR #51, opened 2020-02-12
PricePart currency configuration.

microposmp:
Added a settings option on PricePart that makes it possible to configure/restrict what currencies are possible to select when entering data for a product. This way rules can be setup when the product type is designed and make the entered data more reliable.

**Possible options**:
1. All Currencies - The only option available today. If not configured when attaching the PricePart to a type this is the default.
2. Default Currency - Restricts the PricePart to the configured default currency. Useful for sites that only sells in a single market/region.
3. Specific Currency - The idea with this option is to make it possible to restrict the currency that can be selected for PricePart to a specific currency. Right now this option is disabled.

On a product type multiple PricePart objects can be attached with different configurations.

**Some examples**:

A site that only sells in one market/region can add a single PricePart to the product type and configure it to use the default currency.

A site that targets multiple markets/regions can add more than one PricePart when designing the product type. One PricePart with the default currency and additional PricePart objects configured for a specific currency. For this to work some kind of pricing strategy will need to be configured that selects the price in the correct currency.

image

### DIFF
--- a/Drivers/PricePartDisplayDriver.cs
+++ b/Drivers/PricePartDisplayDriver.cs
@@ -66,15 +66,11 @@ namespace OrchardCore.Commerce.Drivers
 
             switch (part.CurrencySelectionMode)
             {
-                case CurrencySelectionModes.AllCurrencies:
-                    currencySelectionList = _moneyService.Currencies;
-                    break;
-
-                case CurrencySelectionModes.DefaultCurrency:
+                case CurrencySelectionModeEnum.DefaultCurrency:
                     currencySelectionList = new List<ICurrency>() { _moneyService.DefaultCurrency };
                     break;
 
-                case CurrencySelectionModes.SpecificCurrency:
+                case CurrencySelectionModeEnum.SpecificCurrency:
                     currencySelectionList = new List<ICurrency>() { _moneyService.GetCurrency(part.CurrencyIsoCode) };
                     break;
 
--- a/Drivers/PricePartDisplayDriver.cs
+++ b/Drivers/PricePartDisplayDriver.cs
@@ -1,7 +1,10 @@
+using System.Collections.Generic;
 using System.Threading.Tasks;
 using Money;
+using Money.Abstractions;
 using OrchardCore.Commerce.Abstractions;
 using OrchardCore.Commerce.Models;
+using OrchardCore.Commerce.Settings;
 using OrchardCore.Commerce.ViewModels;
 using OrchardCore.ContentManagement.Display.ContentDisplay;
 using OrchardCore.DisplayManagement.ModelBinding;
@@ -51,9 +54,37 @@ namespace OrchardCore.Commerce.Drivers
             model.PriceValue = part.Price.Value;
             model.PriceCurrency = part.Price.Currency == Currency.UnspecifiedCurrency ? _moneyService.DefaultCurrency.CurrencyIsoCode : part.Price.Currency.CurrencyIsoCode;
             model.PricePart = part;
-            model.Currencies = _moneyService.Currencies;
+
+            model.Currencies = GetCurrencySelectionList(part);
 
             return Task.CompletedTask;
+        }
+
+        private IEnumerable<ICurrency> GetCurrencySelectionList(PricePart part)
+        {
+            IEnumerable<ICurrency> currencySelectionList;
+
+            switch (part.CurrencySelectionMode)
+            {
+                case CurrencySelectionModes.AllCurrencies:
+                    currencySelectionList = _moneyService.Currencies;
+                    break;
+
+                case CurrencySelectionModes.DefaultCurrency:
+                    currencySelectionList = new List<ICurrency>() { _moneyService.DefaultCurrency };
+                    break;
+
+                case CurrencySelectionModes.SpecificCurrency:
+                    currencySelectionList = new List<ICurrency>() { _moneyService.GetCurrency(part.CurrencyIsoCode) };
+                    break;
+
+                default:
+                    // As a fallback show all currencies.
+                    currencySelectionList = _moneyService.Currencies;
+                    break;
+            }
+
+            return currencySelectionList;
         }
     }
 }
--- a/Handlers/PricePartHandler.cs
+++ b/Handlers/PricePartHandler.cs
@@ -31,22 +31,20 @@ namespace OrchardCore.Commerce.Handlers
         {
             part.Price = _moneyService.EnsureCurrency(part.Price);
 
-            GetCurrencySelectionMode(part);
-
             return base.LoadingAsync(context, part);
         }
 
         private void GetCurrencySelectionMode(PricePart part)
         {
             var contentTypeDefinition = _contentDefinitionManager.GetTypeDefinition(part.ContentItem.ContentType);
-            var contentTypePartDefinition = contentTypeDefinition.Parts.FirstOrDefault(x => String.Equals(x.PartDefinition.Name, "PricePart"));
+            var contentTypePartDefinition = contentTypeDefinition.Parts.FirstOrDefault(x => x.PartDefinition.Name == nameof(PricePart));
             var currencySelectionMode = contentTypePartDefinition.GetSettings<PricePartSettings>().CurrencySelectionMode;
 
             part.CurrencySelectionMode = currencySelectionMode;
 
-            if (currencySelectionMode == CurrencySelectionModes.SpecificCurrency)
+            if (currencySelectionMode == CurrencySelectionModeEnum.SpecificCurrency)
             {
-                part.CurrencyIsoCode = contentTypePartDefinition.GetSettings<PricePartSettings>().CurrencyIsoCode;
+                part.CurrencyIsoCode = contentTypePartDefinition.GetSettings<PricePartSettings>().SpecificCurrencyIsoCode;
             }
         }
     }
--- a/Handlers/PricePartHandler.cs
+++ b/Handlers/PricePartHandler.cs
@@ -1,23 +1,53 @@
+using System;
+using System.Linq;
 using System.Threading.Tasks;
 using OrchardCore.Commerce.Abstractions;
 using OrchardCore.Commerce.Models;
+using OrchardCore.Commerce.Settings;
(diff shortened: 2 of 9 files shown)


Drivers/PricePartDisplayDriver.cs
```diff
@@ -51,9 +54,37 @@ private Task BuildViewModel(PricePartViewModel model, PricePart part)
...
             return Task.CompletedTask;
         }
+
+        private IEnumerable<ICurrency> GetCurrencySelectionList(PricePart part)
+        {
+            IEnumerable<ICurrency> currencySelectionList;
+
+            switch (part.CurrencySelectionMode)
+            {
+                case CurrencySelectionModes.AllCurrencies:
```
bleroy: Since this is the default, you can remove that case entirely.

Handlers/PricePartHandler.cs
```diff
@@ -1,23 +1,53 @@
...
-        public PricePartHandler(IMoneyService moneyService)
+        public PricePartHandler(IMoneyService moneyService, IContentDefinitionManager contentDefinitionManager)
         {
             _moneyService = moneyService;
+            _contentDefinitionManager = contentDefinitionManager;
+        }
+
+        public override Task InitializingAsync(InitializingContentContext context, PricePart part)
+        {
+            GetCurrencySelectionMode(part);
```
bleroy: Why do we need to do that both on initializing and loading?

Handlers/PricePartHandler.cs
```diff
@@ -1,23 +1,53 @@
...
+
+            GetCurrencySelectionMode(part);
+
             return base.LoadingAsync(context, part);
         }
+
+        private void GetCurrencySelectionMode(PricePart part)
+        {
+            var contentTypeDefinition = _contentDefinitionManager.GetTypeDefinition(part.ContentItem.ContentType);
+            var contentTypePartDefinition = contentTypeDefinition.Parts.FirstOrDefault(x => String.Equals(x.PartDefinition.Name, "PricePart"));
```
bleroy: Style: use `==` and `nameof`.

Settings/PricePartSettings.cs
```diff
@@ -0,0 +1,19 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace OrchardCore.Commerce.Settings
+{
+    public class CurrencySelectionModes
```
bleroy: Why are those strings and not an enum?

Views/PricePartSettings.Edit.cshtml
```diff
@@ -0,0 +1,27 @@
...
+        <div class="input-group">
+            <div class="input-group-append">
+                <select asp-for="CurrencySelectionMode"
+                        asp-items="@(new SelectList(Model.CurrencySelectionModes, "Text", "Value"))"></select>
+            </div>
+        </div>
+
+        <span class="hint">@T["Specifies the currency configuration for this PricePart."]</span>
+
+        @* TODO: MP - Enable the block below when conditional display is done. *@
```
bleroy: I'd prefer to take the PR when it's feature complete.

Handlers/PricePartHandler.cs
microposmp: It wasn't needed in both. Removed from Loading.

Drivers/PricePartDisplayDriver.cs
microposmp: Removed.

Handlers/PricePartHandler.cs
microposmp: Thanks, good advise. Updated.

Settings/PricePartSettings.cs
microposmp: Good point. Changed now. It's still serialized as strings as it makes it easier to read in the json file.

Settings/PricePartSettings.cs
```diff
@@ -0,0 +1,19 @@
...
+    public enum CurrencySelectionModeEnum
+    {
+        AllCurrencies,
+        DefaultCurrency,
+        SpecificCurrency
+    }
+
+    public class PricePartSettings
+    {
+        public string CurrencySelectionMode { get; set; }
```
bleroy: Did you miss this one?

microposmp: I kept this as a string for serialization. That was the reason I made them strings to start with. I'll look at it again and see if I can solve this.

microposmp:
Thanks for the review. It's ok to keep it in this branch until it's complete. Just wanted some feedback on the idea of how to handle this. This replaces the old idea to configure what currencies should be available to select on the site.

Settings/PricePartSettings.cs
bleroy: Serialization of the enum should work fine. I don't think you need to keep this as string.

bleroy:
Yes, I like it better as a setting on the part. Maybe later there can be an additional mode with multi-selection.

Settings/PricePartSettings.cs
microposmp: It's changed to be an enum everywhere now. It's serialized as numeric. I don't get the string conversion to work. It work, but is serialized as a number. Have to look at how the Json is serialized in Orchard Core.

bleroy: If you want the string representation, use the `StringEnumConverter`. `[JsonConverter(typeof(StringEnumConverter))]` on the property.

microposmp: I tried that both with the Newtonsoft and the new System.Text.Json attributes for string convertion. But I didn't get it to work. It's ok to store it as the enum number for now.

Views/PricePartSettings.Edit.cshtml
microposmp: Updated so that the currency selector is shown/hidden based on the CurrencySelectionMode.

Settings/PricePartSettings.cs
bleroy: Where did you put the attribute? (also FWIW, I'll probably remove all the System.Text.Json converters, they were a failed experiment)

microposmp:
Now all review points should be ok. I've thought about multi-selection mode. For that I think we should have a different display in the admin. For example if 3 currencies are selected it should draw 3 entry fields so it will be consistent which currency amount is entered in each separate entry field.

Settings/PricePartSettings.cs
microposmp: I put it on the enum property in PricePartSettings.

microposmp: It works now. [JsonObject] on the class was missing. Have made a new push with this change.

Settings/PricePartSettings.cs
```diff
@@ -0,0 +1,23 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
+
+namespace OrchardCore.Commerce.Settings
+{
+    public enum CurrencySelectionModeEnum
```
bleroy: One last thing, can you move the enum to a separate file?

Views/PricePartSettings.Edit.cshtml
```diff
@@ -0,0 +1,51 @@
...
+        <span class="hint">@T["Currency used for specific currency selection mode."]</span>
+    </fieldset>
+</fieldset>
+
+<script at="Foot">
+$(function () {
+    $('#@Html.IdFor(m => m.CurrencySelectionMode)').change(function () {
+        var value = $('#@Html.IdFor(m => m.CurrencySelectionMode)').val();
+
+        if (value == "SpecificCurrency") {
```
bleroy: Can we use a nameof or something similar so this is more resilient to refactorings?

Views/PricePartSettings.Edit.cshtml
```diff
@@ -0,0 +1,51 @@
...
+    <div class="input-group">
+        <div class="input-group-append">
+            <select asp-for="CurrencySelectionMode"
+                    asp-items="@(new SelectList(Model.CurrencySelectionModes, "Text", "Value"))"></select>
+        </div>
+    </div>
+
+    <span class="hint">@T["Specifies the currency configuration for this PricePart."]</span>
+
+    <fieldset class="input-group" id="speccurr">
```
bleroy: Avoid statically-defined ids: while it's unlikely in this particular situation, these views should be designed so there can be more than one copy on the page. I think you should be able to use a `Html.Id`.

Views/PricePartSettings.Edit.cshtml
```diff
@@ -0,0 +1,51 @@
...
+
+        if (value == "SpecificCurrency") {
+            $('#speccurr').show();
+        } else {
+            $('#speccurr').hide();
+        }
+    });
+});
+
+$(function () {
```
bleroy: Can we avoid the repetition here?

Settings/PricePartSettings.cs
microposmp: Sure.

Views/PricePartSettings.Edit.cshtml
microposmp: Rewrote the javascript code.

Views/PricePartSettings.Edit.cshtml
microposmp: Thanks, great feedback. Implemented it using Html.IdFor.

Views/PricePartSettings.Edit.cshtml
microposmp: Changed the viewmodel to include the editor type.

microposmp:
Thanks for great feedback during the review.

I plan to look at multiple selection mode later. Probably after a pricing strategy is in place that will support selecting the correct currency. It can be achieved today by adding multiple PriceParts to a product type with different currencies. 

For multiple selection mode I think about 2 scenarios:
1. Selecting via the dropdown. Subset of all currencies option available today.
2. A new display mode in admin that will show multiple amount entry fields. One for each currency that is configured in the subset. I think this will provide a better experience for users maintaining products.

bleroy:
I'm partial to columns of checkboxes for this type of multiple choice.

bleroy:
Thanks for the contribution!

PR #54, opened 2020-02-15
Currency selector

microposmp:
Implemented logic to be able to select the currency prices are shown in on the site. It builds on the configuration settings that were added to the PricePart.

**There are two implementations added**: 
1. NullCurrencySelector - Don't affect anything. DefaultCurrency returned from MoneyService will be displayed.
2. CommerceSettingsCurrencySelector - Will use currency configured in commerce settings. Can be different from the default site currency configured. Good for testing.

Update in Startup.cs which implementation will be used.

**Notes**:
- Changed PricePart to be reusable by default.
- Updated PriceProvider to work with multiple currencies on a product. The selected currency is used for showing prices in the shopping cart.
- Added a sample recipe for a product with multiple PriceParts.

### DIFF
--- a/Services/PriceProvider.cs
+++ b/Services/PriceProvider.cs
@@ -4,7 +4,6 @@ using System.Threading.Tasks;
 using OrchardCore.Commerce.Abstractions;
 using OrchardCore.Commerce.Models;
 using OrchardCore.ContentManagement;
-using OrchardCore.ContentManagement.Metadata;
 
 namespace OrchardCore.Commerce.Services
 {
@@ -15,19 +14,13 @@ namespace OrchardCore.Commerce.Services
     {
         private readonly IProductService _productService;
         private readonly IMoneyService _moneyService;
-        private readonly IContentDefinitionManager _contentDefinitionManager;
-        private readonly ITypeActivatorFactory<ContentPart> _contentPartFactory;
 
         public PriceProvider(
             IProductService productService,
-            IMoneyService moneyService,
-            IContentDefinitionManager contentDefinitionManager,
-            ITypeActivatorFactory<ContentPart> contentPartFactory)
+            IMoneyService moneyService)
         {
             _productService = productService;
             _moneyService = moneyService;
-            _contentDefinitionManager = contentDefinitionManager;
-            _contentPartFactory = contentPartFactory;
         }
 
         public int Order => 0;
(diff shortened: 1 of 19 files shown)


Services/PriceProvider.cs
```diff
@@ -30,13 +41,35 @@ public async Task AddPrices(IList<ShoppingCartItem> items)
...
                 if (skuProducts.TryGetValue(item.ProductSku, out var product))
                 {
-                    var pricePart = product.ContentItem.As<PricePart>();
-                    if (pricePart != null)
+                    ContentItem contentItem = product.ContentItem;
+
+                    foreach (var pricePart in PartsOfType<PricePart>(contentItem))
                     {
-                        item.Prices.Add(new PrioritizedPrice(0, pricePart.Price));
+                        if (pricePart.Price.Currency == _moneyService.CurrentDisplayCurrency)
```
bleroy: This is making too strong an assumption on the business rules I think. Specifically, it assumes all products will have a price in the default currency (or something even stronger). I think if the currency selector selects no currency (null), then that means it has no preference, not that the default is the preference.

Actually it's worse than that. The semantics of price providers is that they are additive: their responsibility is to harvest all available prices for this product. Then it's the responsibility of price strategies to select which price should be used in the cart (and in checkout). So this logic should not be here at all.

Services/PriceProvider.cs
```diff
@@ -30,13 +41,35 @@ public async Task AddPrices(IList<ShoppingCartItem> items)
...
                     }
                 }
             }
         }
+
+        public IEnumerable<TPart> PartsOfType<TPart>(ContentItem contentItem) where TPart : ContentPart
+        {
+            var contentTypeDefinition = _contentDefinitionManager.GetTypeDefinition(contentItem.ContentType);
+
+            foreach (var contentTypePartDefinition in contentTypeDefinition.Parts)
```
bleroy: This can be one simple `Select`, `Where` and `Cast<TPart>`

Settings/CommerceSettings.cs
```diff
@@ -9,5 +9,7 @@ public class CommerceSettings
         /// The default currency ISO code
         /// </summary>
         public string DefaultCurrency { get; set; }
+
+        public string CurrentDisplayCurrency { get; set; }
```
bleroy: I think too many of these features are being added to the base feature of the module, adding complexity that not everybody needs. While the base abstractions for all extensibility points should be there, it seems to me like actual multi-currency shops will be the minority, and as such should be isolated in separate features that have to be deliberately activated.

Services/PriceProvider.cs
microposmp: My thinking with the change to add the currency filtering here was to make sure only one currency is returned. The current price strategy has a requirement that only one currency is passed in which I think is a good pattern. If we return all currencies to the shopping cart it will have the responsibility to filter before passing to the price strategies.

The reason the default was used is that most sites will probably only support one currency. If more than one is supported there will most likely be the need to use a currency selector.

Services/PriceProvider.cs
microposmp: This change was to make it work with named parts. I didn't got it to work in any other way.

I had issues with the ContentItem extensions methods named parts.
For example: contentItem.As<PricePart>(); don't return anything for a named part. It only gets a match if the name is the same as the part name.

I'll be doing some more investigation in this area and check with the Orchard Core team.

The original code that was before this didn't work with named parts. I think it's a good reason for adding a test product with this configuration so changes are always tested with it.

Settings/CommerceSettings.cs
microposmp: I agree that this should be a feature. For now I think it's good to have it always available during early development to make it easier to find issues.

microposmp:
I agree that the currency selector should be in it's own feature in the future. The current settings based implementation I think can be removed altogether when a more real world implementation is in place. I think it's good to have right now during early development when concepts are being tried out.

Services/PriceProvider.cs
jeffolmstead: You can retrieve named parts using something like this:
```
contentItem.Get<ContentPart>("MyContentPart")
```
Replace "ContentPart" with the type of part you are retrieving (or keep "ContentPart" if it is a generic part). Obviously, replace "MyContentPart" with the named part "Name".

microposmp: Thanks. I'll try to implement a variant of this that will make it possible to call like the following:
var parts = contentItem.OfType<PricePart>();

I'll do it as an extension method so that the dependency on ContentDefinitionManager can be removed. The idea is to get all parts of a specific type without having to know the name.

microposmp: Simplified the helper method PartsOfType a litte bit. Have also written an extension method for ContentItem which will make it possible to remove the dependency on ContentDefinitionManager and the helper method PartsOfType completely.

```csharp
        /// <summary>
        /// Gets all content elements of a specific type.
        /// </summary>
        /// <typeparam name="TElement">The expected type of the content elements.</typeparam>
        /// <returns>The content element instances or empty sequence if no entries exist.</returns>
        public static IEnumerable<TElement> OfType<TElement>(this ContentElement contentElement) where TElement : ContentElement
        {
            foreach (var part in contentElement.Elements)
            {
                if (part.Value is TElement)
                {
                    var result = part.Value as TElement;

                    yield return result;
                }
            }
        }
```

microposmp:
This is currently broken. Please wait with further review.

Settings/CommerceSettings.cs
bleroy: I think that's going to create problems later. Why not fix it now?

Services/PriceProvider.cs
bleroy: Yeah, I think there's a bug then. I don't see any code doing that selection. Maybe there should be a currency parameter up-front for the whole thing. It's unclear to me how we should select what currency to use when there really are multiple currencies on the site. Need to think more about this.

bleroy: So if you look at the shopping cart controller, you'll see how it groups by currency, but before that, it is possible that prices with multiple currencies may exist on a product and the behavior then is unspecified and probably buggy. I created #56 to track the issue.

Settings/CommerceSettings.cs
microposmp: Ok, I'll make it into a feature.

Startup.cs
```diff
@@ -107,4 +105,14 @@ public override void Configure(IApplicationBuilder app, IEndpointRouteBuilder ro
             );
         }
     }
+
+    // TODO: Move to it's own project? Think about Commerce solution structure...
```
bleroy: I wouldn't.

bleroy:
Is this good to merge now? LGTM.

Startup.cs
microposmp: Removed comment.

microposmp:
It's ok to merge. Thanks.

bleroy:
Thanks!

PR #57, opened 2020-02-21
PriceVariants Part fixes

TRiV07:
Every variant by default have no price.
Determinism issue fix.


Drivers/PriceVariantsPartDisplayDriver.cs
```diff
@@ -12,49 +15,66 @@ namespace OrchardCore.Commerce.Drivers
...
-            model.Currencies = _moneyService.Currencies;
 
-            model.Variants = _priceVariantsService.GetPriceVariants(part.ContentItem);
-            model.VariantsValues = model.Variants.ToDictionary(x => x.Key, x => x.Value.Value);
-            model.VariantsCurrencies = model.Variants.ToDictionary(x => x.Key, x => x.Value.Currency.CurrencyIsoCode);
+            var allVariantsKeys = _predefinedValuesProductAttributeService.GetProductAttributesCombinations(part.ContentItem);
+            model.Variants = part.Variants;
+
+            model.VariantsValues = allVariantsKeys.ToDictionary(x => x,
+                x => model?.Variants?.ContainsKey(x) == true
```
bleroy: You can probably simplify this with a `TryGetValue`, something like `model?.Variants?.TryGetValue(x, out Amount amount) == true ? amount.Value : null` if the compiler lets you. Same thing for currencies. It's weird however that your view model would have two distinct collections for values and currencies. Seems like this could be simpler by using amounts as long as possible?

Services/PredefinedValuesProductAttributeService.cs
```diff
@@ -39,6 +39,7 @@ select accseq.Concat(new[] { item })
         public IEnumerable<ProductAttributeDescription> GetProductAttributesRestrictedToPredefinedValues(ContentItem product)
             => _productAttributeService
                 .GetProductAttributeFields(product)
-                .Where(x => x.Settings is IPredefinedValuesProductAttributeFieldSettings textSettings && textSettings.RestrictToPredefinedValues);
+                .Where(x => x.Settings is IPredefinedValuesProductAttributeFieldSettings textSettings && textSettings.RestrictToPredefinedValues)
+                .OrderBy(x => $"{x.PartName}.{x.Name}");
```
bleroy: Don't format strings just to do a double sort order. Do `OrderBy(x => x.PartName).ThenBy(x => x.Name)` instead.

Services/PriceVariantProvider.cs
```diff
@@ -40,11 +40,12 @@ public async Task AddPrices(IList<ShoppingCartItem> items)
                         {
                             var attributesRestrictedToPredefinedValues = _predefinedValuesService
                                 .GetProductAttributesRestrictedToPredefinedValues(product.ContentItem)
-                                .Select(attr => attr.PartName + "." + attr.Name)
+                                .Select(attr => $"{attr.PartName}.{attr.Name}")
```
bleroy: The concatenation was actually much better.

bleroy:
Thanks for taking care of this!

Drivers/PriceVariantsPartDisplayDriver.cs
TRiV07: we need to have nullable decimal, so best way to use two collections, as for me.

Services/PredefinedValuesProductAttributeService.cs
TRiV07: Fixed

Services/PriceVariantProvider.cs
TRiV07: You're right

bleroy:
Thanks!

PR #60, opened 2020-03-21
Currency serialization

microposmp:
Made a small change to the serialization of PricePart data. Currently data for the unspecified currency is stored. 
```json
	"PricePart": {
		"Price": {
			"value": 999.0,
			"nativename": "Unspecified",
			"englishname": "Unspecified",
			"symbol": "---",
			"iso": "---",
			"currency": "SEK"
		}
	},
```
Added it to the currency table to make it known and serialize like this.
```json
	"PricePart": {
		"Price": {
			"value": 99.0,
			"currency": "SEK"
		}
	},
```

### DIFF
--- a/MoneyDataType/AmountConverter.cs
+++ b/MoneyDataType/AmountConverter.cs
@@ -25,7 +25,6 @@ namespace Money
             string symbol = null;
             string iso = null;
             int? dec = null;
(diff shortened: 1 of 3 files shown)


bleroy:
Thanks!

PR #62, opened 2020-06-26
Not the checkout process

bleroy:
... but some important changes, catch-up with latest Orchard, introducing a shopping cart class, first workflow activity, modeling international addresses, orders, and more. Will merge this sooner rather than later, then continue work towards full checkout process.


jeffolmstead:
Everything I have looked through seems really good. Is there any reason not to merge this? It seems like a stabilizing commit that will serve as a launching point.

Issue #74, opened 2020-11-27
ICurrency == and != operators not returning expected result


DrewBrasher:
I noticed this while stepping through the code in debug mode gathering details for issue #73.  There are 2 places that use the ICurrency == and != operators and I think they should be using .Equals() instead because that method is overridden but the operators cannot be overridden in an interface.

Here are the 2 places the operators are used: https://github.com/OrchardCMS/OrchardCore.Commerce/blob/f6cc5e033b75d8fb6125c80493d17a2ee3471c17/OrchardCore.Commerce/Drivers/PricePartDisplayDriver.cs#L62

https://github.com/OrchardCMS/OrchardCore.Commerce/blob/f6cc5e033b75d8fb6125c80493d17a2ee3471c17/OrchardCore.Commerce/Serialization/PrioritizedPriceConverter.cs#L37

Here is a screenshot from debugging that shows the values:
![image](https://user-images.githubusercontent.com/9272802/100459308-e6c19b00-308a-11eb-8837-def1189b8ca4.png)

and here is a screenshot after changing it to use the .Equals method:
![image](https://user-images.githubusercontent.com/9272802/100459394-05c02d00-308b-11eb-8f96-15ab8aa1f920.png)

PR #75, opened 2020-11-27
Fix for issue #74

DrewBrasher:
Changed instances of ICurrency using == and != operators to use the .Equals() method instead to fix issue #74.


bleroy:
Good catch! Thanks for the fix.

PR #76, opened 2021-01-02
Remove ContentTypes dependency

microposmp:
Made a change to retrieve ProductAttributeField types from the loaded content item.

Removed dependency on OrchardCore.ContentTypes.
Updated to latest Orchard Core version.


bleroy (approved):

bleroy:
Thanks for the contrib!

PR #77, opened 2021-01-02
Currency Test failing in countries with decimal comma (,)

microposmp:
Custom currency without regional/culture configuration fails in countries with decimal comma (,).

Changed test to pass with current system settings. Don't think it will be common use case to add currencies not available in the system.

### DIFF
--- a/OrchardCore.Commerce.Tests/CurrencyTests.cs
+++ b/OrchardCore.Commerce.Tests/CurrencyTests.cs
@@ -18,7 +18,7 @@ namespace OrchardCore.Commerce.Tests
                 { CanadianDollar, 1234.56m, "$1,234.56" },
                 { SwissFranc, 1234.56m, "CHF 1’234.56" },
                 { ChineseYuan, 1234.56m, "¥1,234.56" },
-                { new Currency("My FOO", "My FOO", "f", "FOO"), 1234.56m, "(FOO) 1,234.56" }
+                { new Currency("My FOO", "My FOO", "f", "FOO"), 1234.56m, $"(FOO) {1234.56m.ToString("N")}" }
             };
 
         [Theory]


OrchardCore.Commerce.Tests/CurrencyTests.cs
```diff
@@ -18,7 +18,7 @@ public static CurrencyTheoryData TestData
                 { CanadianDollar, 1234.56m, "$1,234.56" },
                 { SwissFranc, 1234.56m, "CHF 1’234.56" },
                 { ChineseYuan, 1234.56m, "¥1,234.56" },
-                { new Currency("My FOO", "My FOO", "f", "FOO"), 1234.56m, "(FOO) 1,234.56" }
+                { new Currency("My FOO", "My FOO", "f", "FOO"), 1234.56m, $"(FOO) {1234.56m.ToString("N")}" }
```
Skrypt, line 21: I don't think this is appropriate. Here, see he sets an array of fixed values defined with a "Country" or if you prefer "Zone" and then he creates a Currency, which should take any string and assert that it is displaying fine in each "Zone". So here you are forcing a culture currency format on the string to make sure it is compliant with your system but it should not test that, but if it's going to work with ANY culture.

Skrypt: So if you enter "1,234.56" in SwissFranc it should display "CHF 1’234.56".
If your culture doesn't use a comma then it should pass too. If it doesn't then there's an issue in the Currency string parser.

Skrypt, line 21: Not sure, thinking about it.

Skrypt: A null culture normally is not appropriate so it should fallback to InvariantCulture formatting which allows usage of comma for thousands separator.

Skrypt, line 21: Based on this : 

```
        public string ToString(decimal amount)
            => Culture is null
            ? "(" + CurrencyIsoCode + ") " + amount.ToString("N" + DecimalPlaces)
            : amount.ToString("C" + DecimalPlaces, Culture);
```

I think it is appropriate since we do a format of the amount there too.

bleroy:
Good catch, thanks!

PR #84, opened 2021-08-09
New method in IContentManager interface.

microposmp:
Add new member to FakeContentManager. Breaking tests after update to Orchard Core 1.0.


bleroy:
Thanks!

PR #90, opened 2021-11-27
update to dotnet 6

jhgbrt:
updated target framework in all projects
simplified Program.cs and removed Startup.cs 
(as per net5 -> net6 migration guidance)


xperiandri:
I don't think that it is a good idea to remove `Startup.cs`
The simplified template for dummies in .NET 6 is not what a professional Orchard Core project requires.

jhgbrt:
Well, it's a sample intended to demonstrate how OrchardCore Commerce can be used. I would think that especially new developers would expect it to align with the simplified approach (which for better or worse, newcomers will be familiar with in the future).

<opinion>
I believe the bootstrap magic dance between Program and Startup added little to no value. I like the way you now have one place where you can find all bootstrap logic in the new template & philosophy, even for 'advanced' projects.
</opinion>

Of course, I'll leave it to the project maintainers to apply this PR or not.

SampleWebApp/Program.cs
```diff
@@ -1,26 +1,24 @@
...
-        public static IHostBuilder CreateHostBuilder(string[] args) =>
-            Host.CreateDefaultBuilder(args)
-                .ConfigureWebHostDefaults(webBuilder =>
-                {
-                    webBuilder.UseStartup<Startup>();
-                });
-    }
+    app.UseExceptionHandler("/Error");
+    app.UseHsts();
 }
```
Skrypt, line 17: else if (env.IsDevelopment())
            {
                app.UseDeveloperExceptionPage();
            }

SampleWebApp/Program.cs
```diff
@@ -1,26 +1,24 @@
...
+
+app.UseHttpsRedirection();
+app.UseStaticFiles();
+
+app.UseOrchardCore();
+
+app.Run();
```
Skrypt, line 24: new line at eof

Skrypt:
This will probably break if we multi-target frameworks. I think it is better to keep the startup.cs file.

Skrypt (approved):

jhgbrt:
I don't understand your concern. What does the sample web app have to do with multi-targeting?

Skrypt:
We want to be able to test it with the CI by building it with netcoreapp3.1, net5.0, net6.0. And these changes will only work in net6.0 as far as I know?

agriffard:
See https://github.com/OrchardCMS/OrchardCore/issues/10761 for the same concern on the main repo.

Skrypt:
All is good here too. We can merge. I suggest we keep a discussion open about the new way of bootstrapping the app.
Next PR, multi-target.

PR #122, opened 2022-05-15
Add init submodules step to 'Setting up your dev env' in README

davidpuplava:
Cloning this repository using command line doesn't automatically
initiliaze submodules. Running SampleWebApp without submodules
results in a compile error, "file not found
'tools/Lombiq.Analyzers/Build.props'". From README file, it was
not obvious that there were submodules to initialize.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,9 +33,10 @@ Globalization should be taken into account at every step.
 ## Setting up your dev environment
 
 1. Clone this repository.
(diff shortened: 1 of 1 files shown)


agriffard:
Not relevant anymore? 
Should it be closed?

DAud-IcI:
Sorry about the confusion and late reply. It will be removed in the coming days (see #133) so this won't be a problem in the future.

PR #139, opened 2022-08-01
OCC-15: Stop using submodules

### LINKED ISSUE #133: Stop using submodules (OCC-15)
Once [OSOE-110](https://lombiq.atlassian.net/browse/OSOE-110) is resolved we can start using the NuGet version of Lombiq.Analyzers and remove:
- _/tools/Lombiq.Analyzers/_
- /.gitmodules
- the **Submodule Init** step from _/.github/workflows/pr_ci.yml_

DAud-IcI:
OCC-15

Fixes #133


DAud-IcI:
So, the good news is that it works, the bad news is that it fails _per project_. This wasn't obvious during OSOCE-110 where I only tested it on one project at a time. This is only an inconvenience and shouldn't be noticeable except when you introduce analysis to an existing project.

.gitignore
```diff
@@ -14,3 +14,4 @@ node_modules/
 /src/OrchardCore.Commerce.Web/wwwroot/is-cache/
 *.placeholder
 /.editorconfig
+.ps1-analyzer-stamp
```
Piedone, line 17: This should be excluded from the projects (to e.g. not see them in VS).

src/OrchardCore.Commerce.Web/.eslintrc.json
```diff
@@ -0,0 +1,12 @@
+{
+  // Keep in sync with .eslint.global.json.
```
Piedone, line 2: Shouldn't we have such a file? Also, can't we have such a root, global _eslintrc and _stylelintrc_?

src/OrchardCore.Commerce.Web/OrchardCore.Commerce.Web.csproj
```diff
@@ -3,9 +3,15 @@
...
     <CopyRefAssembliesToPublishDirectory>true</CopyRefAssembliesToPublishDirectory>
+
+    <NodeJsExtensionsMarkdownAnalysisMode>solution</NodeJsExtensionsMarkdownAnalysisMode>
+    <ExecDotnetPostcleanCommand>false</ExecDotnetPostcleanCommand>
+    <ExecDotnetPrebuildCommand>false</ExecDotnetPrebuildCommand>
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Lombiq.Analyzers.PowerShell" Version="1.0.0" />
+    <PackageReference Include="Lombiq.NodeJs.Extensions" Version="1.0.0-alpha.17" />
```
Piedone, line 14: Document what one needs to be installed locally to be able to build the solution.

Piedone (changes requested):

src/OrchardCore.Commerce.Web/.eslintrc.json
DAud-IcI: This file just gets added when you reference Lombiq.NoteJs.Extensions. Presumably the _.eslint.global.json_ file exists there. I didn't do anything with _.eslintrc.json_ and _.stylelintrc_. I contemplated adding them to _.gitignore_, but I figured that could cause problems and confusion later if we actually need JS or CSS features in this project.

src/OrchardCore.Commerce.Web/OrchardCore.Commerce.Web.csproj
DAud-IcI: [Apparently nothing?](https://github.com/Lombiq/NodeJs-Extensions/blame/dev/Readme.md#L37)

I guess it helps if you have Node.js installed. :smile_cat: I've added that to the readme. I have just installed a new Linux setup yesterday (Márk suggested a new distro that's right up my alley) so I can confirm that only installing Node (via NVM) is enough.

DAud-IcI: btw this isn't news, the project was [already included in a different module](https://github.com/OrchardCMS/OrchardCore.Commerce/pull/127/files#diff-d1a8d49b5b14a1cdaa27b65272d273762c30616abcfa89fe6f350946c5d9a2bbR14) as it was needed for the Stripe integration. We just didn't think to document it, (probably because pretty much everyone here already has it installed for other projects) so thanks for highlighting that blind spot!

0liver: > a new distro that's right up my alley

@DAud-IcI Which one is it?

DAud-IcI: [EndevourOS](https://endeavouros.com/)

src/OrchardCore.Commerce.Web/.eslintrc.json
Piedone: I see. Can't we have such a root, global _eslintrc_ and _stylelintrc_ though? Or in any case, why just have these for the Web project, when we have assets in `OrchardCore.Commerce`?

DAud-IcI: > Can't we have such a root, global eslintrc and stylelintrc though?

I don't know, ask @0liver. I still don't fully understand the mechanics.

> Or in any case, why just have these for the Web project, when we have assets in `OrchardCore.Commerce`?

[As I said](https://github.com/OrchardCMS/OrchardCore.Commerce/pull/139#discussion_r939727452), `Lombiq.NodeJs.Extensions` was already added to`OrchardCore.Commerce` from OCC-11.
You can see them in `main` [here](https://github.com/OrchardCMS/OrchardCore.Commerce/tree/main/src/Modules/OrchardCore.Commerce#:~:text=5%20days%20ago-,.eslintrc.json,-Temporarily%20make%20CI).

Piedone: Ah, of course. However, why do we need it for the Web project, since there are no assets there? Only for MD linting?

0liver: The comment is somewhat misleading. I've added it only as documentation to the "template" files - they should really not end up in the consuming project.

And they won't after merging from `task/node-extensions` because Lukó moved that comment into a separate _Notes_ file [here](https://github.com/Lombiq/NodeJs-Extensions/pull/8/files#diff-a3555a89ac3d169a81fcb00074256c52af4331693730744c0eda2f9272b3d5acR3).

DAud-IcI: > why do we need it for the Web project, since there are no assets there? Only for MD linting?

Yeah, since you specifically asked to do it this way in OSOCE. :wink: ([also](https://github.com/Lombiq/NodeJs-Extensions/blob/dev/Lombiq.NodeJs.Extensions/Docs/Markdown.md#:~:text=we%20suggest%20setting%20this%20value%20in%20the%20solution's%20%22entry%22%20project%20such%20as%20your%20web%20project))

Piedone: Yeah, I'm aware :). Can we get rid of the rc files though, since we don't need SCSS and JS linting here?

0liver: Set
```xml
<NodeJsExtensionsGlobalESLintConfigurationDirectory></NodeJsExtensionsGlobalESLintConfigurationDirectory>
<NodeJsExtensionsCreateStylelintConfigurationFile></NodeJsExtensionsCreateStylelintConfigurationFile>
```
in the web project's .csproj file.

src/OrchardCore.Commerce.Web/OrchardCore.Commerce.Web.csproj
```diff
@@ -9,8 +9,14 @@
...
 
+  <!-- Currently we don't have any ps1/psm1/psd1 files in the repo, so this is commented out. Add the ProjectReference
+       below when this changes.
   <ItemGroup>
-    <PackageReference Include="Lombiq.Analyzers.PowerShell" Version="1.0.0" />
+    <PackageReference Include="Lombiq.Analyzers.PowerShell" Version="1.0.1-occ-15.0" />
+  </ItemGroup> -->
+
+  <ItemGroup>
+    <PackageReference Include="Lombiq.Analyzers.PowerShell" Version="1.0.1-occ-15.0" />
```
Piedone, line 19: This is left here by mistake, right?

Piedone (changes requested):

src/OrchardCore.Commerce.Web/.eslintrc.json
Piedone: This is missing.

src/OrchardCore.Commerce.Web/OrchardCore.Commerce.Web.csproj
DAud-IcI: Yes.

src/OrchardCore.Commerce.Web/OrchardCore.Commerce.Web.csproj
```diff
@@ -7,6 +7,8 @@
     <NodeJsExtensionsMarkdownAnalysisMode>solution</NodeJsExtensionsMarkdownAnalysisMode>
     <ExecDotnetPostcleanCommand>false</ExecDotnetPostcleanCommand>
     <ExecDotnetPrebuildCommand>false</ExecDotnetPrebuildCommand>
+    <NodeJsExtensionsGlobalESLintConfigurationDirectory></NodeJsExtensionsGlobalESLintConfigurationDirectory>
+    <NodeJsExtensionsCreateStylelintConfigurationFile></NodeJsExtensionsCreateStylelintConfigurationFile>
```
Piedone, line 11: The config files in the repo are now unneeded.

Piedone (changes requested):

src/OrchardCore.Commerce.Web/OrchardCore.Commerce.Web.csproj
DAud-IcI: Hmm, nope they are still created.
image

Piedone: @0liver this shouldn't happen, no?

0liver: Correct - I'd expect them not to be created.

DAud-IcI: Are you sure? I don't think it should work.
Based on [this](
https://github.com/Lombiq/NodeJs-Extensions/blob/fa08be84eac2eeb525a6cb43ed91df81bc60e276/Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.targets#L106-L110) it seems to me that if you set `NodeJsExtensionsGlobalESLintConfigurationDirectory` to empty it means that the path `""` never exists (because it's an invalid path), so the config file is always copied if it doesn't exist.

0liver: You're correct. 😏

DAud-IcI: Well, then this is as far as we can go here.

0liver: Please extend the condition there like so:

```xml
    <Copy SourceFiles="$(NodeJsExtensionsNpmPackageSourcePath)/config/consumer/.eslintrc.project.json"
          DestinationFiles="$(MSBuildProjectDirectory)/.eslintrc.json"
          Condition="
            '$(NodeJsExtensionsGlobalESLintConfigurationDirectory)' != ''
            AND !Exists($(NodeJsExtensionsGlobalESLintConfigurationDirectory))
            AND !Exists('$(MSBuildProjectDirectory)/.eslintrc.json')" />
```

0liver: That was my intention when I wrote that.

src/OrchardCore.Commerce.Web/OrchardCore.Commerce.Web.csproj
```diff
@@ -7,8 +7,8 @@
     <NodeJsExtensionsMarkdownAnalysisMode>solution</NodeJsExtensionsMarkdownAnalysisMode>
     <ExecDotnetPostcleanCommand>false</ExecDotnetPostcleanCommand>
     <ExecDotnetPrebuildCommand>false</ExecDotnetPrebuildCommand>
-    <NodeJsExtensionsGlobalESLintConfigurationDirectory></NodeJsExtensionsGlobalESLintConfigurationDirectory>
-    <NodeJsExtensionsCreateStylelintConfigurationFile></NodeJsExtensionsCreateStylelintConfigurationFile>
+    <NodeJsExtensionsGlobalESLintConfigurationDirectory>false</NodeJsExtensionsGlobalESLintConfigurationDirectory>
+    <NodeJsExtensionsCreateStylelintConfigurationFile>false</NodeJsExtensionsCreateStylelintConfigurationFile>
```
DAud-IcI, line 11: This is also changed to false according to https://github.com/Lombiq/NodeJs-Extensions/blob/dev/Lombiq.NodeJs.Extensions/Docs/Styles.md?plain=1#L77

PR #140, opened 2022-08-02
OCC-17: Inconsistent currency configuration after setup

### LINKED ISSUE #105: Inconsistent currency configuration after setup (OCC-17)
While trying out the module I was surprised to see it says "Can't add product sample because it doesn't have a price." when trying to add it to the cart, even though I have correctly set up a USD price:
image

When I go to /Admin/Settings/commerce both currency options are shown to be "USD", but when I place a breakpoint in `MoneyService.CurrentDisplayCurrency` it actually says the value is `---`:
image
(By the way, since `_currencySelector.CurrentDisplayCurrency` is not `null` by default I don't think the `?? DefaultCurrency` is ever used.)

Looks like the frontend works correctly but the admin config gets the wrong default. If I click the Save button in the Commerce settings without touching anything it gets updated to USD and starts to behave as expected.

DemeSzabolcs:
OCC-17
Fixes #105

### DIFF
--- a/src/Modules/OrchardCore.Commerce/Services/ShoppingCartHelpers.cs
+++ b/src/Modules/OrchardCore.Commerce/Services/ShoppingCartHelpers.cs
@@ -83,7 +83,10 @@ public class ShoppingCartHelpers : IShoppingCartHelpers
         {
             await _notifier.AddAsync(
                 NotifyType.Error,
-                T["Can't add product {0} because it doesn't have a price.", line.ProductSku]);
+                T[
+                    "Can't add product {0} because it doesn't have a price, or its currency doesn't match with the " +
+                    "current display currency.",
+                    line.ProductSku]);
             return null;
         }
 
--- a/src/Modules/OrchardCore.Commerce/Services/ShoppingCartHelpers.cs
+++ b/src/Modules/OrchardCore.Commerce/Services/ShoppingCartHelpers.cs
@@ -83,7 +83,10 @@ public class ShoppingCartHelpers : IShoppingCartHelpers
         {
             await _notifier.AddAsync(
                 NotifyType.Error,
+                // Localized strings need to be in one line.
+#pragma warning disable S103 // Long line (which is greater than 150 authorized)
                 T["Can't add product {0} because it doesn't have a price, or its currency doesn't match with the current display currency.", line.ProductSku]);
+#pragma warning restore S103 // Long line (which is greater than 150 authorized)
             return null;
         }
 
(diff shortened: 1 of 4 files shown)


src/Modules/OrchardCore.Commerce/Services/ShoppingCartHelpers.cs
```diff
@@ -83,7 +83,10 @@ public async Task<ShoppingCartItem> ParseCartLineAsync(ShoppingCartLineUpdateMod
         {
             await _notifier.AddAsync(
                 NotifyType.Error,
-                T["Can't add product {0} because it doesn't have a price.", line.ProductSku]);
+                T[
+                    "Can't add product {0} because it doesn't have a price, or its currency doesn't match with the " +
+                    "current display currency.",
+                    line.ProductSku]);
```
DAud-IcI, line 89: The localization string must be on the same line, otherwise [the tooling](https://docs.orchardcore.net/en/dev/docs/reference/modules/Localize/#extract-translations-to-po-files) won't recognize it.

image

src/Modules/OrchardCore.Commerce/Views/PricePart.cshtml
```diff
@@ -1,6 +1,26 @@
 @model PricePartViewModel
 
-@if (Model.CurrentDisplayCurrency.CurrencyIsoCode == Model.PriceCurrency)
-{
-<div class="product-price">@Model.Price.ToString()</div>
+@{
+    var currentDisplayCurrencyIsoCode = Model.CurrentDisplayCurrency.CurrencyIsoCode;
+    var modelPriceCurrency = Model.PriceCurrency;
+    var modelPrice = Model.Price.ToString();
```
DAud-IcI, line 6: do you actually need the ToString? I think that's what both usages should do anyway.

DAud-IcI (changes requested):

src/Modules/OrchardCore.Commerce/Services/ShoppingCartHelpers.cs
```diff
@@ -83,7 +83,10 @@ public async Task<ShoppingCartItem> ParseCartLineAsync(ShoppingCartLineUpdateMod
         {
             await _notifier.AddAsync(
                 NotifyType.Error,
-                T["Can't add product {0} because it doesn't have a price.", line.ProductSku]);
+                // Localized strings need to be in one line.
+#pragma warning disable S103 // Long line (which is greater than 150 authorized)
+                T["Can't add product {0} because it doesn't have a price, or its currency doesn't match with the current display currency.", line.ProductSku]);
```
0liver, line 88: I know I'm too late for the party, but `match` sounds better without `with`:

```suggestion
                T["Can't add product {0} because it doesn't have a price, or its currency doesn't match the current display currency.", line.ProductSku]);
```

Same in the view.

PR #146, opened 2022-08-09
OCC-24: Product with PriceVariantsPart can't be added to the cart

### LINKED ISSUE #143: Product with PriceVariantsPart can't be added to the cart (OCC-24)
Because `PriceProvider` is called first and then it gets NRE.

I suggest adding a method to `IPriceProvider`, something like:
```csharp
bool IsApplicable(IList<ShoppingCartItem> items)
```
Then call it with `.Where()` everywhere that `.OrderBy(provider => provider.Order)` is used so only the valid ones will be evaluated. (You should still keep the `foreach` because someone may add multiple applicable providers for their own use-cases.)

DemeSzabolcs:
OCC-24

Fixes #143


src/Modules/OrchardCore.Commerce/Services/PriceProvider.cs
```diff
@@ -27,8 +28,8 @@ public class PriceProvider : IPriceProvider
 
     public async Task<IEnumerable<ShoppingCartItem>> AddPricesAsync(IList<ShoppingCartItem> items)
     {
-        var skus = items.Select(item => item.ProductSku).Distinct().ToArray();
-        var skuProducts = (await _productService.GetProductsAsync(skus)).ToDictionary(productPart => productPart.Sku);
+        var skuProducts = await PriceProviderHelpers.GetSkuProductsAsync(items, _productService);
```
0liver, line 31: If you transform that helper method into an extension on `IProductService`, this line will stay more idiomatic:
```suggestion
        var skuProducts = await _productService.GetSkuProductsAsync(items);
```

src/Modules/OrchardCore.Commerce/Services/PriceProvider.cs
```diff
@@ -55,6 +56,22 @@ public async Task<ShoppingCartItem> AddPriceAsync(ShoppingCartItem item)
...
+        return items
+            .All(item =>
+            {
+                if (skuProducts.TryGetValue(item.ProductSku, out var productPart))
+                {
+                    return productPart.ContentItem.OfType<PricePart>().Any();
+                }
+
+                return false;
+            });
```
0liver, line 72: How about saving a few lines?
```suggestion
        return items.Select(item => skuProducts.TryGetValue(item.ProductSku, out var productPart)
            ? AddPriceToShoppingCartItem(item, productPart)
            : item);
```

src/Modules/OrchardCore.Commerce/Services/PriceProvider.cs
```diff
@@ -27,8 +28,8 @@ public class PriceProvider : IPriceProvider
 
     public async Task<IEnumerable<ShoppingCartItem>> AddPricesAsync(IList<ShoppingCartItem> items)
     {
-        var skus = items.Select(item => item.ProductSku).Distinct().ToArray();
-        var skuProducts = (await _productService.GetProductsAsync(skus)).ToDictionary(productPart => productPart.Sku);
+        var skuProducts = await PriceProviderHelpers.GetSkuProductsAsync(items, _productService);
+
         return items
             .Select(item =>
             {
```
0liver, line 35: This return statement can be shortened to this:

```c#
        return items.Select(item => skuProducts.TryGetValue(item.ProductSku, out var productPart)
            ? AddPriceToShoppingCartItem(item, productPart)
            : item);
```

src/Modules/OrchardCore.Commerce/Services/PriceVariantProvider.cs
```diff
@@ -94,4 +94,21 @@ private ShoppingCartItem AddPriceToShoppingCartItem(ShoppingCartItem item, Produ
...
+            .All(item =>
+            {
+                if (skuProducts.TryGetValue(item.ProductSku, out var productPart))
+                {
+                    var contentItem = productPart.ContentItem;
+                    return contentItem.OfType<PricePart>().Any() || contentItem.OfType<PriceVariantsPart>().Any();
+                }
+
+                return false;
+            });
```
0liver, line 112: This can be shortened along [these lines](https://github.com/OrchardCMS/OrchardCore.Commerce/pull/146/files#r941884894), too.

src/Modules/OrchardCore.Commerce/Services/PriceService.cs
```diff
@@ -17,7 +17,11 @@ public class PriceService : IPriceService
 
     public async Task<IList<ShoppingCartItem>> AddPricesAsync(IList<ShoppingCartItem> items)
     {
-        foreach (var priceProvider in _providers.OrderBy(provider => provider.Order))
+        var providers = await _providers
+            .OrderBy(provider => provider.Order)
+            .WhereAsync(provider => provider.IsApplicableAsync(items));
```
DAud-IcI, line 22: You forgot to do the same for `AddPriceAsync` here:
https://github.com/OrchardCMS/OrchardCore.Commerce/blob/b671bae20827a73cb288d571d5d4fbc17f82fc7f/src/Modules/OrchardCore.Commerce/Services/PriceService.cs#L35

Also can you explain why we needed separate `AddPriceAsync` method at all? Seems to me that it should be just an extension method for `IPriceService` that returns something like `(await priceService.AddPricesAsync(new [] { item })).Single()` instead of a full implementation. Then we wouldn't need `IPriceProvider.AddPriceAsync()` either.

src/Modules/OrchardCore.Commerce/Services/PriceVariantProvider.cs
```diff
@@ -94,4 +94,14 @@ private ShoppingCartItem AddPriceToShoppingCartItem(ShoppingCartItem item, Produ
...
     }
+
+    public async Task<bool> IsApplicableAsync(IList<ShoppingCartItem> items)
+    {
+        var skuProducts = await _productService.GetSkuProductsAsync(items);
+
+        return items
+            .All(item => (skuProducts.TryGetValue(item.ProductSku, out var productPart) &&
+                productPart.ContentItem.OfType<PricePart>().Any()) ||
+                productPart.ContentItem.OfType<PriceVariantsPart>().Any());
```
DAud-IcI, line 105: I don't understand this. Why are you checking for `PricePart` and why is it applicable if you have it? I think `PriceVariantProvider` is only relevant if you have `PriceVariantsPart`. A content type shouldn't have both parts anyway.

Also you put the parenthesis on the wrong place so we have this:
image

src/Modules/OrchardCore.Commerce/Services/ProductService.cs
```diff
@@ -31,7 +35,19 @@ public async Task<IEnumerable<ProductPart>> GetProductsAsync(IEnumerable<string>
...
             .Distinct()
             .ToArray();
-        return (await _contentManager.GetAsync(contentItemIds))
-            .Select(item => item.As<ProductPart>());
+
+        var contentItems = await _contentManager.GetAsync(contentItemIds);
+
+        // We need BuildDisplayAsync to fill part.Elements with the fields. We could extract the logic from
+        // BuildDisplayAsync, but we would have to copy almost everything from there.
+        if (contentItems.Any())
```
DAud-IcI, line 43: Why is checking `.Any()` necessary? If `contentItems` is empty the `foreach` will just skip anyway.

DAud-IcI (changes requested):
Cal you also add a unit test? Just to open a product and try adding it to the cart. You can copy the recipe from [OCC-18](https://lombiq.atlassian.net/browse/OCC-18).

src/Modules/OrchardCore.Commerce/Services/PriceVariantProvider.cs
DemeSzabolcs: `PriceVariantProvider` in my experience can handle content items with `PricePart` too that's why I added both, so it's technically "applicable", but you are right.

src/Modules/OrchardCore.Commerce/Services/PriceService.cs
DemeSzabolcs: I added this back then when I did the Stripe implementation. I will try your suggestion.

src/Modules/OrchardCore.Commerce/Recipes/OrchardCore.Commerce.Development.Tests.Setup.recipe.json
```diff
@@ -19,6 +19,10 @@
         {
           "executionid": "OrchardCore.Commerce",
           "name": "OrchardCore.Commerce.Development.Setup"
+        },
+        {
+          "executionid": "OrchardCore.Commerce",
+          "name": "TestProduct"
```
DemeSzabolcs, line 25: The `PriceVariantsProduct` was OK for testing, but it didn't work, since it didn't have `ProductAttributeField`. I moved it to a different recipe and created a new one, that has a `ProductAttributeField`.

src/Modules/OrchardCore.Commerce/Recipes/TestProduct.recipe.json
```diff
@@ -0,0 +1,58 @@
...
+              "Versionable": true,
+              "Securable": true
+            }
+          },
+          "ContentTypePartDefinitionRecords": [
+            {
+              "PartName": "TestPriceVariantsProduct",
+              "Name": "TestPriceVariantsProduct",
+              "Settings": {}
+            },
```
DAud-IcI, line 42: What's this for?

src/Modules/OrchardCore.Commerce/Services/ProductService.cs
```diff
@@ -37,11 +38,42 @@ public async Task<IEnumerable<ProductPart>> GetProductsAsync(IEnumerable<string>
 
         var contentItems = await _contentManager.GetAsync(contentItemIds);
 
-        // We need BuildDisplayAsync to fill part.Elements with the fields. We could extract the logic from
-        // BuildDisplayAsync, but we would have to copy almost everything from there.
+        // We need to replicate some things that BuildDisplayAsync does to fill part.Elements with the fields. We can't
```
DAud-IcI, line 41:
```suggestion
        // We have to replicate some things that BuildDisplayAsync does to fill part.Elements with the fields. We can't
```

src/Modules/OrchardCore.Commerce/Services/ProductService.cs
```diff
@@ -37,11 +38,42 @@ public async Task<IEnumerable<ProductPart>> GetProductsAsync(IEnumerable<string>
...
         foreach (var contentItem in contentItems)
         {
-            await _contentDisplayHandler.BuildDisplayAsync(contentItem, context: null);
+            var contentItemsPartName = contentItem.ContentType;
+
+            var contentItemsPartFields = _contentDefinitionManager.GetTypeDefinition(contentItem.ContentType).Parts
+                .FirstOrDefault(contentPartDefinition =>
+                    contentPartDefinition.PartDefinition.Name == contentItemsPartName)
+                .PartDefinition
+                .Fields;
```
DAud-IcI, line 51:
```suggestion
            var contentItemsPartFields = _contentDefinitionManager
                .GetTypeDefinition(contentItem.ContentType)
                .Parts
                .FirstOrDefault(contentPartDefinition =>
                    contentPartDefinition.PartDefinition.Name == contentItemsPartName)
                .PartDefinition
                .Fields;
```

src/Modules/OrchardCore.Commerce/Services/ProductService.cs
```diff
@@ -37,11 +38,42 @@ public async Task<IEnumerable<ProductPart>> GetProductsAsync(IEnumerable<string>
...
+        // We need to replicate some things that BuildDisplayAsync does to fill part.Elements with the fields. We can't
+        // use BuildDisplayAsync directly because it requires a BuildDisplayContext.
         foreach (var contentItem in contentItems)
         {
-            await _contentDisplayHandler.BuildDisplayAsync(contentItem, context: null);
+            var contentItemsPartName = contentItem.ContentType;
+
+            var contentItemsPartFields = _contentDefinitionManager.GetTypeDefinition(contentItem.ContentType).Parts
+                .FirstOrDefault(contentPartDefinition =>
+                    contentPartDefinition.PartDefinition.Name == contentItemsPartName)
```
DAud-IcI, line 49: `.FirstOrDefault` can return `null`. Did you mean `.First`?

src/Modules/OrchardCore.Commerce/Services/ProductService.cs
```diff
@@ -37,11 +38,42 @@ public async Task<IEnumerable<ProductPart>> GetProductsAsync(IEnumerable<string>
...
         var contentItems = await _contentManager.GetAsync(contentItemIds);
 
-        // We need BuildDisplayAsync to fill part.Elements with the fields. We could extract the logic from
-        // BuildDisplayAsync, but we would have to copy almost everything from there.
+        // We need to replicate some things that BuildDisplayAsync does to fill part.Elements with the fields. We can't
+        // use BuildDisplayAsync directly because it requires a BuildDisplayContext.
         foreach (var contentItem in contentItems)
         {
-            await _contentDisplayHandler.BuildDisplayAsync(contentItem, context: null);
+            var contentItemsPartName = contentItem.ContentType;
```
DAud-IcI, line 45: Seems to me that this will only work for the content types created in the Admin UI. Any custom content type is unlikely to have a content type named part. (e.g. Product content type would have a part named ProductPart rather than Product, so this wouldn't find it)

I think you should get every field from all parts, and then filter to only include those that inherit from the `ProductAttributeField` abstract class.

src/Modules/OrchardCore.Commerce/Services/ProductService.cs
```diff
@@ -37,11 +38,42 @@ public async Task<IEnumerable<ProductPart>> GetProductsAsync(IEnumerable<string>
...
+                        break;
+
+                    case nameof(TextProductAttributeField):
+                        part.Get<TextProductAttributeField>(fieldName);
+                        break;
+
+                    default:
+                        continue;
+                }
+            }
```
DAud-IcI, line 76: What is this for? You don't do anything with the results.

DAud-IcI (changes requested):

src/Modules/OrchardCore.Commerce/Services/ProductService.cs
DemeSzabolcs: Yes, but this is what makes it work (fill the fields in the content items). I looked at the source code and in that, this is what fills the fields.

DemeSzabolcs: I can add a comment about it, but this part is essential.

DAud-IcI: Strange, but okay. Anyway, we will certainly add more product attribute types in the future and then it will break. It should be done for every type that derives from `ProductAttributeField`.

DemeSzabolcs: The problem is that I can't really check that, since the types of the fields are only given in `string`s. In the original code it uses generics.

Original code:
```
contentPart.Get<TField>(partFieldDefinition.Name);
```
I tried to use get with `ProductAttributeField` but it wants to create a C# object out of it and `ProductAttributeField` is an abstract class:
```
Type is an interface or abstract class and cannot be instantiated. Path 'PriceVariantsProduct.Size'.
```
A solution to `Get` all the fields (not relying on the fact that there are three `ProductAttributeField`s now) could be this:
```
part.Get(Type.GetType("OrchardCore.Commerce.Fields." + field.FieldDefinition.Name), fieldName);
```
What do you think?

DemeSzabolcs: Maybe I could also get `"OrchardCore.Commerce.Fields."` dynamically, but I think that doesn't change.

PR #152, opened 2022-08-16
Documentation

agriffard:
Add MkDocs


agriffard:
Preview: https://orchardcorecommerce.readthedocs.io/en/ag-docs/

Let me know what you think of it and which kind of doc structure we could adopt.
Ex: 
- Getting Started.
- 1 Readme per feature / library.

agriffard:
Then, I will be able to create a sub domain like commerce.orchardcore.net pointing to the docs site.

.readthedocs.yml
```diff
@@ -0,0 +1,22 @@
...
+  configuration: mkdocs.yml
+
+# Optionally build your docs in additional formats such as PDF and ePub
+formats: all
+
+# Optionally set the version of Python and requirements required to build your docs
+python:
+  version: 3.7
+  install:
+    - requirements: src/docs/requirements.txt
```
DAud-IcI, line 22: Add trailing newline to get rid of the :no_entry: here, please. Same for 404.html and main.html.

.readthedocs.yml
```diff
@@ -0,0 +1,22 @@
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+# Build documentation in the docs/ directory with Sphinx
+#sphinx:
+#  configuration: docs/conf.py
+
```
DAud-IcI, line 10: This can be deleted.

mkdocs.yml
```diff
@@ -0,0 +1,85 @@
+site_name: Orchard Core Commerce
+theme:
+  name: material
+  custom_dir: src/docs/theme
```
DAud-IcI, line 4: Can we make these directory names inside `src/` capitalized?
To be consistent with the others like Modules and Libraries.

src/README.md
```diff
@@ -0,0 +1,76 @@
+# Orchard Core Commerce
```
DAud-IcI, line 1: Did you copy the _README.md_ from the repository root? Why? This will be a maintenance problem.

src/docs/OrchardCore.Commerce.Docs.csproj
```diff
@@ -0,0 +1,8 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
DAud-IcI, line 1: Could you explain why we need the csproj file?
Nothing in it will be built, but it will create a useless NuGet package once  #151 is merged.

src/docs/releases/0.0.1.md
```diff
@@ -0,0 +1,3 @@
+# Orchard Core Commerce 0.0.1
```
DAud-IcI, line 1: Is this file auto-generated? If yes, how do we update _releases_ in the future? Please link a resource in the root readme.

DAud-IcI (changes requested):

DAud-IcI:
Thanks, the preview looks nice. Should we also have a workflow for updating this after every merge to `main`? (does OC have an existing workflow we can steal? :slightly_smiling_face:)

I agree with "1 Readme per feature / library.", incidentally I have already added stub readmes for the two existing libraries in #151 , e.g. [here](https://github.com/OrchardCMS/OrchardCore.Commerce/pull/151/files#diff-6e58ff03da6ed3008ced1b1892625432657ee02844cbf93115bf6c2bb9b9e426). For Getting Started is fine in the main readme, but later we can turn it into a separate page as you suggest.

src/docs/OrchardCore.Commerce.Docs.csproj
agriffard: > Could you explain why we need the csproj file? 
To make it appear in the solution.

> Nothing in it will be built, but it will create a useless NuGet package once #151 is merged.
It shouldn't be built because of the `<IsPackable>false</IsPackable>`

src/docs/releases/0.0.1.md
agriffard: Manually created, as in OC.
Creating a release in GitHub will generate the PR merged since the latest release.
After first release (starting at the second), a link to compare the two releases could be included.

Change in .md, .yml files should not trigger the build.

We can document the release process as in https://docs.orchardcore.net/en/latest/docs/topics/publishing-releases/

src/README.md
agriffard: Yes, I copied it as in OC because it becomes the root of the documentation.

Next goals will be to:
- Split the infos in multiple pages.
- Make the root Readme lighter. Ex: Just description, badges and link to doc site.

.readthedocs.yml
agriffard: line endings added

mkdocs.yml
agriffard: Sorry, I won't change this for the moment but I am thinking about moving the src\docs to the root:
- It is only placed there so we can make links to code file (which we rarely do).
- It would solve your problem with the duplicated Readme (at root and in src).

DAud-IcI: Moving docs to root sounds good.

src/docs/OrchardCore.Commerce.Docs.csproj
DAud-IcI: Okay then.

src/docs/releases/0.0.1.md
DAud-IcI: I see. I'm sure we'll streamline this after the MVP is done. Until then I don't anticipate new releases (this v0.0.1 was just to test the NuGet publishing workflow) so we can leave it as-is.

mkdocs.yml
agriffard: docs moved to root.

src/README.md
agriffard: We need to keep docs/README.md, it will be the first page of the docs site.

DAud-IcI: Please don't resolve the conversations yourself, it makes reviewing harder as I have to look through the closed convos.

Check out [this issue](https://github.com/mkdocs/mkdocs/issues/2831#issuecomment-1204114853), looks like you can include files outside of the docs directory using the [Snippets extension](https://facelessuser.github.io/pymdown-extensions/extensions/snippets/#snippets-notation). Please try that for _docs/README.md_. I also suggest renaming it to _docs/OrchardCore.Commerce.md_ so when we include readmes for the other libraries (or any future modules) it will be consistent.

agriffard: Snippet works but we can't refer to the root Readme.

DAud-IcI: Have you read the specific comment that I've linked in the issue? It explains where you have to put the redirect: inside the md file, not in the mkdocs.yml. You should have a file like this: https://github.com/pawamoy/duty/blob/master/docs/index.md

DAud-IcI (changes requested):
Please merge and see my comment [here](https://github.com/OrchardCMS/OrchardCore.Commerce/pull/152#discussion_r950020033).

src/README.md
agriffard: Ok, done. It seems to work:
https://orchardcorecommerce.readthedocs.io/en/ag-docs/

agriffard: `Error: MD041: first-line-heading: First line in a file should be a top-level heading. Rule information: https://github.com/DavidAnson/markdownlint/blob/v0.26.2/doc/Rules.md#md041`

DAud-IcI: Great! I've [disabled](https://github.com/DavidAnson/markdownlint#configuration) markdownlint for the file.

agriffard: F... da rules.

src/docs/OrchardCore.Commerce.Docs.csproj
DAud-IcI: Unfortunately I found out after merge that this breaks `dotnet pack`, so I have to delete the csproj file after all.
see https://github.com/OrchardCMS/OrchardCore.Commerce/runs/7962504898?check_suite_focus=true#step:4:4202

PR #161, opened 2022-08-21
OCC-32: Add Checkout permission

### LINKED ISSUE #154: Add Checkout permission (OCC-32)
Check the permission in `PaymentController.Index` (you get there after clicking the <kbd>Checkout</kbd> button). If the authorization succeeds continue loading the page. Otherwise:
- if the client is authenticated, return `Forbidden`.
- if the client is anonymous, redirect to the login screen with ReturnUrl being `~/checkout`.

DemeSzabolcs:
OCC-32

Fixes #154

### DIFF
--- a/src/Modules/OrchardCore.Commerce/Drivers/StripeApiSettingsDisplayDriver.cs
+++ b/src/Modules/OrchardCore.Commerce/Drivers/StripeApiSettingsDisplayDriver.cs
@@ -22,20 +22,20 @@ public class StripeApiSettingsDisplayDriver : SectionDisplayDriver<ISite, Stripe
     private readonly IDataProtectionProvider _dataProtectionProvider;
     private readonly IShellHost _shellHost;
     private readonly ShellSettings _shellSettings;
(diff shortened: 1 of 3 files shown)


src/Modules/OrchardCore.Commerce/Drivers/StripeApiSettingsDisplayDriver.cs
```diff
@@ -22,28 +22,28 @@ public class StripeApiSettingsDisplayDriver : SectionDisplayDriver<ISite, Stripe
     private readonly IDataProtectionProvider _dataProtectionProvider;
     private readonly IShellHost _shellHost;
     private readonly ShellSettings _shellSettings;
-    private readonly IHttpContextAccessor _httpContextAccessor;
+    private readonly IHttpContextAccessor _hca;
```
DemeSzabolcs, line 25: This is the name what the Lombiq Dependency injection VS extension suggests, so I replaced it here too.

DAud-IcI (changes requested):

PR #174, opened 2022-09-08
OCC-28: Order part breaks when it contains products with attributes

### LINKED ISSUE #148: Order part breaks when it contains products with attributes (OCC-28)
### Summary

Order part breaks when it contains products with attributes. Meaning when opening an order content item, the order part itself will not be displayed.

### Reproduction steps
**Some of the reproduction steps require #146 to be closed, or the usage of the [OCC-24](https://github.com/OrchardCMS/OrchardCore.Commerce/tree/issue/OCC-24) branch.** 

1. Configure Stripe payment (see [Readme](https://github.com/OrchardCMS/OrchardCore.Commerce/blob/main/README.md))
2. Create a content type with `Product` and `PriceVariants` parts, add and configure a `TextProductAttributeField`. (Alternatively use the [OCC-24](https://github.com/OrchardCMS/OrchardCore.Commerce/tree/issue/OCC-24) branch or wait for it to be merged since it contains a content type and item that I previously mentioned)
3. Create a content item of the previously created type.
4. Add it to the shopping cart, then checkout and "pay" (see [Readme](https://github.com/OrchardCMS/OrchardCore.Commerce/blob/main/README.md) again for test cards). To be able to add products with `PriceVariants` you need to use the [OCC-24](https://github.com/OrchardCMS/OrchardCore.Commerce/tree/issue/OCC-24) branch or wait for #146 to be closed.
5. Go to the content items and click _Edit_ on the newly created order.
6. Observe that the `OrderPart` is missing.

### Results
Expected: The `OrderPart` is visible, the attributes are visible on the products inside the order.
Actual: The `OrderPart` is not visible.

image

### Additional notes
A log file is created when trying to open an order. Important part of it:
```
2022-08-11 15:52:11.7555|Default|00-eedd49b989fb09108df32f3e11138bbf-bdec484b7983fea8-00||OrchardCore.ContentManagement.DefaultContentManager|ERROR|IContentHandler thrown from
OrchardCore.ContentManagement.Handlers.ContentPartHandlerCoordinator by JsonSerializationException Newtonsoft.Json.JsonSerializationException: Could not create an instance of type
OrchardCore.Commerce.Abstractions.IProductAttributeValue. Type is an interface or abstract class and cannot be instantiated. Path 'OrderPart.LineItems[0].Attributes[0].UntypedPredefinedValue'.
```

DemeSzabolcs:
OCC-28

Fixes #148

### DIFF
--- a/src/Modules/OrchardCore.Commerce/Serialization/ProductAttributeValueConverter.cs
+++ b/src/Modules/OrchardCore.Commerce/Serialization/ProductAttributeValueConverter.cs
@@ -1,0 +1,75 @@
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using OrchardCore.Commerce.Abstractions;
+using OrchardCore.Commerce.ProductAttributeValues;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace OrchardCore.Commerce.Serialization;
+
+internal class ProductAttributeValueConverter : JsonConverter<IProductAttributeValue>
+{
+    private const string Type = "type";
+    private const string Value = "value";
+    private const string AttributeName = "attributename";
+
+    public override IProductAttributeValue ReadJson(
+        JsonReader reader,
+        Type objectType,
+        IProductAttributeValue existingValue,
+        bool hasExistingValue,
+        JsonSerializer serializer)
+    {
+        IDictionary<string, JToken> propertiesAndValues = new Dictionary<string, JToken>();
+        while (reader.Read() && reader.TokenType == JsonToken.PropertyName)
+        {
+            var propertyName = reader.Value.ToString();
+            reader.Read();
+            var propertyValue = JToken.ReadFrom(reader);
+
+            propertiesAndValues.Add(propertyName, propertyValue);
+        }
+
+        var attributeName = propertiesAndValues[AttributeName].ToString();
+
+        return propertiesAndValues[Type].ToString() switch
+        {
+            nameof(TextProductAttributeValue) => new TextProductAttributeValue(attributeName, propertiesAndValues[Value].ToString()),
+            nameof(BooleanProductAttributeValue) => new BooleanProductAttributeValue(attributeName, propertiesAndValues[Value].ToObject<bool>()),
+            nameof(NumericProductAttributeValue) => new NumericProductAttributeValue(attributeName, propertiesAndValues[Value].ToObject<decimal>()),
+            _ => null,
+        };
+    }
+
+    public override void WriteJson(JsonWriter writer, IProductAttributeValue productAttributeValue, JsonSerializer serializer)
+    {
+        if (productAttributeValue is null)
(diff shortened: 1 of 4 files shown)


src/Modules/OrchardCore.Commerce/Serialization/ProductAttributeValueConverter.cs
```diff
@@ -0,0 +1,74 @@
...
+using System;
+using System.Collections.Generic;
+
+namespace OrchardCore.Commerce.Serialization;
+
+internal class ProductAttributeValueConverter : JsonConverter<IProductAttributeValue>
+{
+    private const string Type = "type";
+    private const string Value = "value";
+    private const string AttributeName = "attributename";
```
DAud-IcI, line 14:
```suggestion
    private const string AttributeName = "attributeName";
```

src/Modules/OrchardCore.Commerce/Serialization/ProductAttributeValueConverter.cs
```diff
@@ -0,0 +1,74 @@
...
+            _ => throw new InvalidOperationException($"Unknown or unsupported type \"{typeName}\"."),
+        };
+    }
+
+    public override void WriteJson(JsonWriter writer, IProductAttributeValue productAttributeValue, JsonSerializer serializer)
+    {
+        if (productAttributeValue is null)
+        {
+            return;
+        }
```
DAud-IcI, line 42:
```suggestion
        if (productAttributeValue is null) return;
```

src/Modules/OrchardCore.Commerce/Serialization/ProductAttributeValueConverter.cs
```diff
@@ -0,0 +1,74 @@
...
+        writer.WriteStartObject();
+
+        var typeName = productAttributeValue.GetType().Name;
+
+        writer.WritePropertyName(Type);
+        writer.WriteValue(typeName);
+
+        writer.WritePropertyName(Value);
+
+        if (typeName == nameof(TextProductAttributeValue))
```
DAud-IcI, line 53: Instead of hard coding the type, it would be cleaner to check if `productAttributeValue.UntypedValue` is `IEnumerable<string>`, or even just `IEnumerable`.

DAud-IcI (changes requested):

src/Modules/OrchardCore.Commerce/Serialization/ProductAttributeValueConverter.cs
0liver: I've hit an unexpected case doing the same in CARE:

I was checking a property for its type and value, and when the type was derived from `IEnumerable` and the value was `null` I threw an `ArgumentError` asking the reader to assign an empty collection to the property when construction its containing class.

Turns out I forgot that `string : IEnumerable<char>`.

You might want to account for that, too.

PR #194, opened 2022-09-28
OCC-88: Add site setting to limit the list of available countries/regions.

### LINKED ISSUE #187: Add site setting to limit the list of available countries/regions. (OCC-88)

When the business doesn't want to ship globally.

- Add a site setting and an editor for it where you can list the allowed country codes. If it's empty then `Regions.All` should be used, as before.
- Add an `IRegionService` with a method to look up the list of available regions.
- Find all usages of `Regions.All` and and replace them with the aforementioned service.

[Jira issue](https://lombiq.atlassian.net/browse/OCC-88)

DemeSzabolcs:
OCC-88

Fixes #187

### DIFF
--- a/src/Modules/OrchardCore.Commerce/Abstractions/IRegionService.cs
+++ b/src/Modules/OrchardCore.Commerce/Abstractions/IRegionService.cs
@@ -1,0 +1,17 @@
+using System.Collections.Generic;
+using System.Globalization;
+using System.Threading.Tasks;
+
+namespace OrchardCore.Commerce.Abstractions;
+
+/// <summary>
+/// When implemented a service for handling regions.
+/// </summary>
+public interface IRegionService
+{
+    /// <summary>
+    /// Gets the available regions from the site settings./>.
+    /// </summary>
+    /// <returns>A collection of <see cref="RegionInfo"/>.</returns>
+    Task<IEnumerable<RegionInfo>> GetAvailableRegionsAsync();
+}
--- a/src/Modules/OrchardCore.Commerce/Abstractions/IRegionService.cs
+++ b/src/Modules/OrchardCore.Commerce/Abstractions/IRegionService.cs
@@ -10,7 +10,7 @@ namespace OrchardCore.Commerce.Abstractions;
 public interface IRegionService
 {
     /// <summary>
-    /// Gets the available regions from the site settings./>.
+    /// Gets the available regions from the site settings.
     /// </summary>
     /// <returns>A collection of <see cref="RegionInfo"/>.</returns>
     Task<IEnumerable<RegionInfo>> GetAvailableRegionsAsync();
--- a/src/Modules/OrchardCore.Commerce/Abstractions/IRegionService.cs
+++ b/src/Modules/OrchardCore.Commerce/Abstractions/IRegionService.cs
@@ -5,7 +5,7 @@ using System.Threading.Tasks;
 namespace OrchardCore.Commerce.Abstractions;
 
 /// <summary>
-/// When implemented a service for handling regions.
+/// A service for accessing and customizing regions.
 /// </summary>
 public interface IRegionService
 {
--- a/src/Modules/OrchardCore.Commerce/AdminMenu.cs
+++ b/src/Modules/OrchardCore.Commerce/AdminMenu.cs
@@ -31,6 +31,10 @@ public class AdminMenu : INavigationProvider
                     .Add(T["Stripe API"], T["Stripe API"], stripeApi => stripeApi
                     .Action("Index", "Admin", new { area = "OrchardCore.Settings", groupId = StripeApiSettingsDisplayDriver.GroupId })
                     .Permission(Permissions.ManageStripeApiSettings)
+                    .LocalNav())
+                    .Add(T["Region"], T["Region"], region => region
+                    .Action("Index", "Admin", new { area = "OrchardCore.Settings", groupId = RegionSettingsDisplayDriver.GroupId })
+                    .Permission(Permissions.ManageRegionSettings)
                     .LocalNav())));
 
         return Task.CompletedTask;
--- a/src/Modules/OrchardCore.Commerce/AdminMenu.cs
+++ b/src/Modules/OrchardCore.Commerce/AdminMenu.cs
@@ -29,13 +29,21 @@ public class AdminMenu : INavigationProvider
                         .Permission(Permissions.ManageCommerceSettings)
                         .LocalNav())
                     .Add(T["Stripe API"], T["Stripe API"], stripeApi => stripeApi
-                    .Action("Index", "Admin", new { area = "OrchardCore.Settings", groupId = StripeApiSettingsDisplayDriver.GroupId })
-                    .Permission(Permissions.ManageStripeApiSettings)
-                    .LocalNav())
+                        .Action("Index", "Admin", new
+                        {
+                            area = "OrchardCore.Settings",
+                            groupId = StripeApiSettingsDisplayDriver.GroupId,
+                        })
+                        .Permission(Permissions.ManageStripeApiSettings)
+                        .LocalNav())
                     .Add(T["Region"], T["Region"], region => region
-                    .Action("Index", "Admin", new { area = "OrchardCore.Settings", groupId = RegionSettingsDisplayDriver.GroupId })
-                    .Permission(Permissions.ManageRegionSettings)
-                    .LocalNav())));
+                        .Action("Index", "Admin", new
+                        {
+                            area = "OrchardCore.Settings",
+                            groupId = RegionSettingsDisplayDriver.GroupId,
+                        })
+                        .Permission(Permissions.ManageRegionSettings)
+                        .LocalNav())));
 
         return Task.CompletedTask;
     }
--- a/src/Modules/OrchardCore.Commerce/Drivers/RegionSettingsDisplayDriver.cs
+++ b/src/Modules/OrchardCore.Commerce/Drivers/RegionSettingsDisplayDriver.cs
@@ -1,3 +1,4 @@
+using Dapper;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.DataProtection;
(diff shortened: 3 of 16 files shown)


src/Modules/OrchardCore.Commerce/Abstractions/IRegionService.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace OrchardCore.Commerce.Abstractions;
+
+/// <summary>
+/// When implemented a service for handling regions.
+/// </summary>
+public interface IRegionService
+{
+    /// <summary>
+    /// Gets the available regions from the site settings./>.
```
DAud-IcI, line 13: typo?
```suggestion
    /// Gets the available regions from the site settings.
```

src/Modules/OrchardCore.Commerce/Abstractions/IRegionService.cs
```diff
@@ -0,0 +1,17 @@
+using System.Collections.Generic;
+using System.Globalization;
+using System.Threading.Tasks;
+
+namespace OrchardCore.Commerce.Abstractions;
+
+/// <summary>
+/// When implemented a service for handling regions.
```
DAud-IcI, line 8: "When implemented" is redundant for an interface. It's terminology usually used in abstract types to indicate that implementation depends on the child. I know I have used it in `IPaymentProvider` too, that was also wrong of me.

```suggestion
/// A service for accessing and customizing regions.
```

src/Modules/OrchardCore.Commerce/AdminMenu.cs
```diff
@@ -31,6 +31,10 @@ public Task BuildNavigationAsync(string name, NavigationBuilder builder)
                     .Add(T["Stripe API"], T["Stripe API"], stripeApi => stripeApi
                     .Action("Index", "Admin", new { area = "OrchardCore.Settings", groupId = StripeApiSettingsDisplayDriver.GroupId })
                     .Permission(Permissions.ManageStripeApiSettings)
+                    .LocalNav())
+                    .Add(T["Region"], T["Region"], region => region
```
DAud-IcI, line 35: Indentation seems to be missing for this and the "Stripe API" above.

src/Modules/OrchardCore.Commerce/Drivers/RegionSettingsDisplayDriver.cs
```diff
@@ -0,0 +1,91 @@
...
+using OrchardCore.Environment.Shell;
+using OrchardCore.Settings;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace OrchardCore.Commerce.Drivers;
+
+public class RegionSettingsDisplayDriver : SectionDisplayDriver<ISite, RegionSettings>
+{
+    public const string GroupId = "Region";
```
DAud-IcI, line 21:
```suggestion
    public const string GroupId = "Region";

```

src/Modules/OrchardCore.Commerce/Drivers/RegionSettingsDisplayDriver.cs
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    public override async Task<IDisplayResult> EditAsync(RegionSettings section, BuildEditorContext context)
+    {
+        var user = _hca.HttpContext?.User;
+
+        if (!await _authorizationService.AuthorizeAsync(user, Permissions.ManageRegionSettings))
+        {
+            return null;
+        }
```
DAud-IcI, line 48: It fits in a line.
```suggestion
        if (!await _authorizationService.AuthorizeAsync(user, Permissions.ManageRegionSettings)) return null;
```

src/Modules/OrchardCore.Commerce/Drivers/RegionSettingsDisplayDriver.cs
```diff
@@ -0,0 +1,91 @@
...
+            .OnGroup(GroupId);
+    }
+
+    public override async Task<IDisplayResult> UpdateAsync(RegionSettings section, BuildEditorContext context)
+    {
+        var user = _hca.HttpContext?.User;
+
+        if (!await _authorizationService.AuthorizeAsync(user, Permissions.ManageRegionSettings))
+        {
+            return null;
```
DAud-IcI, line 66: Shouldn't this explicitly throw instead of return `null`? I think you'd get an exception either way, but that would be better for any future bug tracking in AI.

src/Modules/OrchardCore.Commerce/Drivers/RegionSettingsDisplayDriver.cs
```diff
@@ -0,0 +1,91 @@
...
+        {
+            var model = new RegionSettingsViewModel();
+
+            if (await context.Updater.TryUpdateModelAsync(model, Prefix))
+            {
+                var allowedRegions = model.AllowedRegions;
+
+                var allRegionTwoLetterIsoRegionNames = Regions.All.Select(region => region.TwoLetterISORegionName);
+
+                section.AllowedRegions = allowedRegions != null && allowedRegions.Any()
```
DAud-IcI, line 79: You can combine the two with null-conditional operator and then just check if it's `true`

src/Modules/OrchardCore.Commerce/Drivers/RegionSettingsDisplayDriver.cs
```diff
@@ -0,0 +1,91 @@
...
+            return null;
+        }
+
+        if (context.GroupId == GroupId)
+        {
+            var model = new RegionSettingsViewModel();
+
+            if (await context.Updater.TryUpdateModelAsync(model, Prefix))
+            {
+                var allowedRegions = model.AllowedRegions;
```
DAud-IcI, line 75: This is `IEnumerable<string>` and you have multiple enumerations below (both the `.Any()` check and that `.Contains()` is called N times). Add `.AsList()` to enforce that it's a type that can be safely multiple enumerated.

src/Modules/OrchardCore.Commerce/Services/RegionConfiguration.cs
```diff
@@ -0,0 +1,25 @@
...
+
+    public void Configure(RegionSettings options)
+    {
+        var settings = _siteService
+            .GetSiteSettingsAsync()
+            .GetAwaiter()
+            .GetResult()
+            .As<RegionSettings>();
+
+        options.AllowedRegions = settings.AllowedRegions;
```
DAud-IcI, line 23: Can you create an recipe that initializes this to 3-5 countries? It should not be included in the setup recipe but available for unit testing.

DAud-IcI (changes requested):

DAud-IcI:
Also can you add a unit test checks if the country list is limited? You can use the new Order content item editor to verify the country list the easiest.

src/Modules/OrchardCore.Commerce/Drivers/RegionSettingsDisplayDriver.cs
DemeSzabolcs: I copied this from `StripeApiSettingsDisplayDriver.cs` should I change it there too?

DemeSzabolcs: It was also `null` [in OC](https://github.com/OrchardCMS/OrchardCore/blob/main/src/OrchardCore.Modules/OrchardCore.ReCaptcha/Drivers/ReCaptchaSectionDisplayDriver.cs#L38-L41), that's why I used `null` here too.

DemeSzabolcs: And if I should throw, then what? `UnauthorizedAccessException`?

src/Modules/OrchardCore.Commerce/Services/RegionConfiguration.cs
DemeSzabolcs: I also made region settings exportable.

test/OrchardCore.Commerce.Tests.UI/Tests/RegionTests/PersistenceRegionSettingsTests.cs
```diff
@@ -0,0 +1,41 @@
...
+        ExecuteTestAfterSetupAsync(
+            async context =>
+            {
+                await context.SignInDirectlyAsync();
+                await context.ExecuteRecipeDirectlyAsync("RegionTest");
+
+                await context.CreateNewContentItemAsync("Order");
+
+                var selectListOptions = context
+                    .GetAll(By.XPath("//select[@id='OrderPart_BillingAddress_Address_Region']/option"));
```
DAud-IcI, line 30: Use the `id` function in XPath: `id('OrderPart_BillingAddress_Address_Region')/option`

test/OrchardCore.Commerce.Tests.UI/Tests/RegionTests/PersistenceRegionSettingsTests.cs
```diff
@@ -0,0 +1,41 @@
...
+
+                selectListOptions.Count.ShouldBe(3);
+
+                var selectListTexts = selectListOptions.Select(selectListOption => selectListOption.Text).AsList();
+
+                selectListTexts[0].ShouldBe("Argentina");
+                selectListTexts[1].ShouldBe("Luxemburg");
+                selectListTexts[2].ShouldBe("Magyarország");
```
DAud-IcI, line 38: This whole thing can be simplified if you check for array like `context.getAll(...).Select(...).ToArray().SholdBe(new [] { ... });`

DAud-IcI (changes requested):

src/Modules/OrchardCore.Commerce/Drivers/RegionSettingsDisplayDriver.cs
DAud-IcI: If it's the same in OC then it can say.

PR #199, opened 2022-10-05
OCC-23: Product variants should have unique SKUs

### LINKED ISSUE #135: Product variants should have unique SKUs (OCC-23)
Suggested by @chinasqzl [here](https://github.com/OrchardCMS/OrchardCore.Commerce/issues/5#issuecomment-1016995456).

This could be handled similarly to variant prices. However inventory management may complicate things. An approach where variants are separate content items and they are linked (similar to how the Content Localization feature works) may be more appropriate. Please discuss.

porgabi:
OCC-23

Fixes #135


src/Modules/OrchardCore.Commerce/Abstractions/IProductService.cs
```diff
@@ -14,12 +15,25 @@ public interface IProductService
...
+    Task<(PriceVariantsPart Part, string VariantKey)> GetExactVariantAsync(string sku);
 }
 
 public static class ProductServiceExtensions
 {
-    public static async Task<ProductPart> GetProductAsync(this IProductService service, string sku) =>
-        (await service.GetProductsAsync(new[] { sku })).FirstOrDefault();
+    public static async Task<ProductPart> GetProductAsync(this IProductService service, string sku)
+    {
+        var dashIndex = sku.IndexOf(value: "-", StringComparison.InvariantCulture);
```
DemeSzabolcs, line 34: `sku.IndexOf(value: "-", StringComparison.InvariantCulture);` is also in `ProductService.cs`. Please put it in an extension method and use that.

src/Modules/OrchardCore.Commerce/Services/ProductService.cs
```diff
@@ -76,4 +76,18 @@ private static void FillField(ContentPart part, ContentPartFieldDefinition field
             part.Get(typeOfField, fieldName);
         }
     }
+
+    public string GetVariantKey(string sku)
```
DemeSzabolcs, line 80: Sorry, I just noticed something. `productPart.Sku + '-' + value.Classify().ToUpperInvariant` This part is missing.

> Oliver is right, you only save the base SKU. The variant SKU should be generated similarly to [how the price variant labels are generated](https://github.com/OrchardCMS/OrchardCore.Commerce/blob/main/src/Modules/OrchardCore.Commerce/Drivers/PriceVariantsPartDisplayDriver.cs#L82) and then converted to that upper case format I previously described.

DemeSzabolcs (changes requested):

src/Modules/OrchardCore.Commerce/Abstractions/IProductService.cs
```diff
@@ -29,9 +29,11 @@ public interface IProductService
 
 public static class ProductServiceExtensions
 {
+    public static int GetDashIndex(string sku) => sku.IndexOf(value: "-", StringComparison.InvariantCulture);
```
DemeSzabolcs, line 32: This is not an extension method.

Do it like this:
```suggestion
    public static int GetDashIndex(this string sku) => sku.IndexOf(value: "-", StringComparison.InvariantCulture);
```
Then you can use it like this
```
var dashIndex = sku.GetDashIndex();
```

Also could you please move the whole `ProductServiceExtensions` class to a different file to the `Extensions` folder. It's getting ugly that it's here.

src/Modules/OrchardCore.Commerce/Services/ProductService.cs
```diff
@@ -79,14 +79,14 @@ private static void FillField(ContentPart part, ContentPartFieldDefinition field
 
     public string GetVariantKey(string sku)
     {
-        var dashIndex = sku.IndexOf(value: "-", StringComparison.InvariantCulture);
-        return dashIndex == -1 ? sku : sku[(dashIndex + 1)..];
+        var dashIndex = ProductServiceExtensions.GetDashIndex(sku);
+        return dashIndex == -1 ? sku : sku[(dashIndex + 1)..].ToUpperInvariant();
```
DemeSzabolcs, line 83: This is still wrong. Let me clarify, this function is okay (apart from `.ToUpperInvariant()` see below), I just commented here, because there is no specific place where I can comment this.

`productPart.Sku + '-' + value.Classify().ToUpperInvariant` is still missing. It's okay that we can give this function an SKU and a variant key and it will get the variant key, but the variant key is generated nowhere. There is no function to currently get the SKU + variant key. 

Like if we have a size L, black shirt, and its base SKU is SHIRT, there is no way to get SHIRT-L-BLACK (so the SKU and  the variant key). 

Dávid mentioned this here (please read this carefully, also see the linked code): https://github.com/OrchardCMS/OrchardCore.Commerce/issues/135#issuecomment-1251175671

You don't need to save it, but there should be a way to generate it. It might not be used currently, but it should be implemented. (I think you also asked this:

> Makes sense; just to clarify, these display surfaces are to be implemented in the future and do not yet exist, right?

)
Though please check if anything need to be replaced by the SKU + variant key combination where `PriceVariantPart` is used for a product.

Also, you don't need to convert it to `ToUpperVariant()`, the base SKU will be uppercase anyway and if you create the above-mentioned variant key generator function you can attach the variant key there as uppercase, so it will "arrive" here as uppercase already.

@DAud-IcI Could you please confirm that I'm right? Since the idea was yours.

DemeSzabolcs (changes requested):

src/Modules/OrchardCore.Commerce/Abstractions/IProductService.cs
DAud-IcI: The `comparisonType` should be `StringComparison.Ordinal`. When you compare punctuation this is the simplest and preferred type.
Also I don't think this needs to be an extension method because using split would be better in both use cases.

src/Modules/OrchardCore.Commerce/Abstractions/IProductService.cs
```diff
@@ -14,12 +15,27 @@ public interface IProductService
...
 public static class ProductServiceExtensions
 {
-    public static async Task<ProductPart> GetProductAsync(this IProductService service, string sku) =>
-        (await service.GetProductsAsync(new[] { sku })).FirstOrDefault();
+    public static int GetDashIndex(string sku) => sku.IndexOf(value: "-", StringComparison.InvariantCulture);
+
+    public static async Task<ProductPart> GetProductAsync(this IProductService service, string sku)
+    {
+        var dashIndex = GetDashIndex(sku);
+        return (await service.GetProductsAsync(new[] { dashIndex == -1 ? sku : sku[..dashIndex] })).FirstOrDefault();
```
DAud-IcI, line 37: Use `.SingleOrDefault`, because it should never return more than one results and if it does then that's an exception.

src/Modules/OrchardCore.Commerce/Services/ProductService.cs
```diff
@@ -76,4 +76,18 @@ private static void FillField(ContentPart part, ContentPartFieldDefinition field
...
+
+    public string GetVariantKey(string sku)
+    {
+        var dashIndex = ProductServiceExtensions.GetDashIndex(sku);
+        return dashIndex == -1 ? sku : sku[(dashIndex + 1)..].ToUpperInvariant();
+    }
+
+    public async Task<(PriceVariantsPart Part, string VariantKey)> GetExactVariantAsync(string sku)
+    {
+        var productPart = await ProductServiceExtensions.GetProductAsync(this, sku);
```
DAud-IcI, line 88: Why not `await this.GetProductAsync(sku)`?

src/Modules/OrchardCore.Commerce/Services/ProductService.cs
DAud-IcI: The problem is that currently there is no relation between the SKU fragment and the `PriceVariantsPart.Variants` dictionary key. The goal is to be able to look up the specific variant price from the variant SKU. The best approach would be to update [here](https://github.com/OrchardCMS/OrchardCore.Commerce/blob/2ed945f5996a71cba81b7b17ab61e1f86713ba53/src/Modules/OrchardCore.Commerce/Drivers/PriceVariantsPartDisplayDriver.cs#L81) and [here](https://github.com/OrchardCMS/OrchardCore.Commerce/blob/6645257c7dd248e30ee915c6527524bb4d657131/src/Modules/OrchardCore.Commerce/Services/PriceVariantProvider.cs#L65), basically uppercase and classify the dictionary key to get the SKU style we want. Then we won't have to convert back and forth between two different key types.

Once updated, please extract [this section](https://github.com/OrchardCMS/OrchardCore.Commerce/blob/6645257c7dd248e30ee915c6527524bb4d657131/src/Modules/OrchardCore.Commerce/Services/PriceVariantProvider.cs#L60-L69) into a method of `ShoppingCartItem` because it will be reused by future features. Also create a unit test to verify that the SKU variant key is generated correctly from the `ShoppingCartItem` attributes.

I agree that you don't need `ToUpperInvariant()`, but on the other hand please add `StringComparer.OrdinalIgnoreCase` to the Dictionary constructor [here](https://github.com/OrchardCMS/OrchardCore.Commerce/blob/a4408afb203b0e5a135a19c38f18fac2ab0f4efe/src/Modules/OrchardCore.Commerce/Models/PriceVariantsPart.cs#L12) to avoid backwards compatibility issues with existing instances.

src/Modules/OrchardCore.Commerce/Abstractions/IProductService.cs
DAud-IcI, line 37: This is a nice method, but removing the part after the dash should be in the `GetProductsAsync` implementation. Otherwise you could pass variant SKUs into `GetProductsAsync` and that would be invalid.

Also I think it would be cleaner to do `sku.Split('-').First()` instead of slicing.

src/Modules/OrchardCore.Commerce/Services/ProductService.cs
```diff
@@ -76,4 +76,18 @@ private static void FillField(ContentPart part, ContentPartFieldDefinition field
             part.Get(typeOfField, fieldName);
         }
     }
+
+    public string GetVariantKey(string sku)
+    {
+        var dashIndex = ProductServiceExtensions.GetDashIndex(sku);
+        return dashIndex == -1 ? sku : sku[(dashIndex + 1)..].ToUpperInvariant();
```
DAud-IcI, line 83: Some time ago I've made a method for splitting a string into two parts, called `Partition` (name and concept shamelessly stolen from Python). I think if you don't have a dash here then something went wrong and it should throw.
So this better written as `sku.Partition("-").Right ?? throw new ArgumentException("The SKU doesn't contain a dash. Is it a product variant SKU?", nameof(sku))`.

DAud-IcI (changes requested):

src/Modules/OrchardCore.Commerce/Services/ProductService.cs
porgabi: After adding uppercase conversion and classification to the highlighted parts in **PriceVariantsPartDisplayDriver** and **PriceVariantProvider**, it's not clear to me if a standalone SKU-generator method is necessary?

Primary reason I've been slow to address this part is I've still been struggling to understand how exactly it's supposed to work. Please let me know if my latest additions are correct.

DAud-IcI: It's okay. I wanted a standalone method for future-proofing, but we can do that when the need actually arises.

DAud-IcI (changes requested):

PR #200, opened 2022-10-05
SITE-29: Make separate editor for checkout

DemeSzabolcs:
SITE-29


src/Modules/OrchardCore.Commerce/Views/Payment/Index.cshtml
```diff
@@ -1,16 +1,269 @@
...
+            <div class="mb-3">
+                <label asp-for="OrderPart.Phone.Text">@T["Phone Number"]</label>
+                <input asp-for="OrderPart.Phone.Text" class="form-control content-preview-text" required />
+            </div>
+            <div class="address_billing-address">
+                <h3>@T["Billing Address"]</h3>
+
+                <div class="row">
+                    <div class="mb-3 col-md-12">
+                        <label asp-for="OrderPart.BillingAddress.Address.Name">@T["Name"]</label>
```
DAud-IcI, line 32: Can't you display the _AddressField_Edit.cshtml_ using a `<shape>` instead of copy-pasting the template? If that's not possible, move the shared part of _AddressField_Edit.cshtml_ into a reusable shape and use that in all 3 places.

src/Modules/OrchardCore.Commerce/Views/Payment/Index.cshtml
```diff
@@ -1,16 +1,269 @@
...
+            </div>
+
+            <div class="mb-3">
+                <label for="StripePaymentPart_PaymentMethodId_Text">PaymentMethodId</label>
+                <input class="form-control content-preview-text"
+                       dir="ltr"
+                       type="text"
+                       id="StripePaymentPart_PaymentMethodId_Text"
+                       name="StripePaymentPart.PaymentMethodId.Text">
+            </div>
```
DAud-IcI, line 262: Any Stripe-specific inputs should be in the `StripeCheckout` shape.
Also you don't need `<div>` and `<label>`, just an `<input hidden id="..." name="...">`.

DAud-IcI (changes requested):

src/Modules/OrchardCore.Commerce/Views/Payment/Index.cshtml
DemeSzabolcs: > Can't you display the AddressField_Edit.cshtml 
The problem with that, is that I have only an `OrderPart`, I can get the order part's addresses, but that will be an `AddressField` not and `AddressFieldViewModel`.

I'm not sure that I can do that with all 3 of them. I can do it for the checkout page for sure. By passing the `OrderPart`, and a parameter that tells which one we want to display (shipping or billing address). But for the _AddressField_Edit_ it's problematic, since I can't pass the `OrderPart` from there, also it might be not on an `OrderPart`. 

So my main problems here are the `asp-for` values. Also I can't just pass an `Address` because there will be no link to the `OrderPart` or any other part (it will be just Address_Name etc.).

DAud-IcI: You can use `id` and `name` attributes in the shape instead of `asp-for`. (Where the id is just the name but dots are replaced with underscores.) The shape can get the individual properties or a new `AddressFieldViewModel` property, you can easily make two if those in the _Index.cshtml_. Then include an additional property, e.g. `CityName` (which gets `@Html.NameFor(model => model.OrderPart.BillingAddress.Address.City)`). You'd only need this for one field (City is chosen arbitrarily because it's distinct and short) as it's only used to generate the name prefix. They you get the prefixes in the shape like this:
```csharp
@{
    var namePrefix = ((string)Model.CityName)[..^(nameof(AddressFieldViewModel.City).Length)]
    var idPrefix = namePrefix.Replace('.', '_');
}
...
<label for="@(idPrefix + nameof(AddressFieldViewModel.Name))">@T["Name"]</label>
...
<input id="@(idPrefix + nameof(AddressFieldViewModel.Name))" name="@(namePrefix + nameof(AddressFieldViewModel.Name))" class="form-control content-preview-text address__name" required />
```

DAud-IcI: Also now that I think about it, the whole content of _Payment/Index.cshtml_ should be extracted into a shape too, otherwise users won't be able to customize the screen from DotNest.

src/Modules/OrchardCore.Commerce/Views/Payment/Index.cshtml
DAud-IcI: Now you don't need [these SCSS rules](https://github.com/OrchardCMS/OrchardCore.Commerce/blob/a52468420ee40d12f633d03667b1ba84a5a4802e/src/Modules/OrchardCore.Commerce/Assets/Styles/credit-card-form.scss#L28-L36) either.

src/Modules/OrchardCore.Commerce/Controllers/PaymentController.cs
```diff
@@ -133,6 +142,27 @@ public async Task<IActionResult> SuccessPost(string orderId)
         if (await _contentManager.GetAsync(orderId) is not { } order) return NotFound();
 
         await _contentItemDisplayManager.UpdateEditorAsync(order, _updateModelAccessor.ModelUpdater, isNew: false);
+
+        // Saving addresses.
+        var userService = _userServiceLazy.Value;
+        var orderPart = order.As<OrderPart>();
+
+        if (await userService.GetCurrentFullUserAsync(_hca) is { } user)
```
DAud-IcI, line 150: You don't need to inject HCA in a controller, use `userService.GetFullUserAsync(User)`.

DAud-IcI (changes requested):

PR #208, opened 2022-10-17
OCC-97: Country or region selector wrong spelling

### LINKED ISSUE #204: Country or region selector wrong spelling (OCC-97)
### Describe the bug

In the AddressField the "Country or region" selector is displaying Netherlands with incorrect spelling as Nederlân. There might be others, this was pointed out by a Dutch.

### To Reproduce

Most straightforward steps to reproduce the behavior:

1. Go to Region settings /Admin/Settings/Region
4. See the error

### Expected behavior

The correct way of spelling for every country. I see that this is coming from the System RegionInfo. I would suggest to use the `EnglishName` instead of the `DisplayName` in the `RegionExtensions`' `CreateSelectListOptions` function. Not only because this way it is correctly spelled, but it is more like the generic way in webshops to list countries.

[Jira issue](https://lombiq.atlassian.net/browse/OCC-97)

DAud-IcI:
OCC-97
Fixes #204

### DIFF
--- a/src/Libraries/OrchardCore.Commerce.AddressDataType/Regions.cs
+++ b/src/Libraries/OrchardCore.Commerce.AddressDataType/Regions.cs
@@ -15,8 +15,14 @@ public static class Regions
         CultureInfo
             .GetCultures(CultureTypes.SpecificCultures)
             .Where(culture => !culture.Equals(CultureInfo.InvariantCulture) && !culture.IsNeutralCulture)
-            .Select(culture => new RegionInfo(culture.LCID))
+            .Select(culture =>
+            {
+                // This sometimes throws "CultureNotFoundException: Culture is not supported." exception on Linux only.
+                try { return new RegionInfo(culture.LCID); }
+                catch (CultureNotFoundException) { return null; }
+            })
             .Where(region =>
+                region != null &&
                 region.TwoLetterISORegionName.Length == 2 && // Filter out world and other 3-digit regions.
                 !string.IsNullOrEmpty(region.EnglishName))
             .Distinct()
--- a/src/Libraries/OrchardCore.Commerce.AddressDataType/Regions.cs
+++ b/src/Libraries/OrchardCore.Commerce.AddressDataType/Regions.cs
@@ -10,7 +10,7 @@ public static class Regions
     /// <summary>
     /// Gets the list of regions.
     /// </summary>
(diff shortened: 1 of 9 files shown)


src/Modules/OrchardCore.Commerce/Extensions/RegionExtensions.cs
```diff
@@ -1,18 +1,14 @@
...
 
 namespace OrchardCore.Commerce.Extensions;
 
 public static class RegionExtensions
 {
-    public static IEnumerable<SelectListItem> CreateSelectListOptions(this IEnumerable<RegionInfo> regionInfos) =>
-        regionInfos.OrderBy(region => region.DisplayName).Select(region => new SelectListItem(
-            region.DisplayName,
+    public static IEnumerable<SelectListItem> CreateSelectListOptions(this IEnumerable<Region> regionInfos) =>
+        regionInfos.OrderBy(region => region.EnglishName).Select(region => new SelectListItem(
```
DemeSzabolcs, line 11: I think `regionInfos` should be renamed to `regions` if you use the name `region` inside the `OrderBy` and `Select`.

src/Libraries/OrchardCore.Commerce.AddressDataType/Regions.cs
```diff
@@ -10,27 +10,25 @@ public static class Regions
     /// <summary>
     /// Gets the list of regions.
     /// </summary>
-    public static IList<RegionInfo> All { get; } =
+    [Obsolete("Use service! This is a temporary attribute.")]
+    public static IList<Region> All { get; } =
         CultureInfo
             .GetCultures(CultureTypes.SpecificCultures)
+            .Where(culture => !culture.Equals(CultureInfo.InvariantCulture) && !culture.IsNeutralCulture)
```
DemeSzabolcs, line 17: Hmm. I think this is not needed, since we are using `.GetCultures(CultureTypes.SpecificCultures)` instead of `.GetCultures(CultureTypes.AllCultures)`. I checked the results of `.GetCultures(CultureTypes.SpecificCultures)` and I couldn't find any cultures with `InvariantCulture`, or `IsNeutralCulture`.

Also with and without this `Where` condition, `Regions.All` have 135 members anyway.

DemeSzabolcs (changes requested):

PR #216, opened 2022-10-19
Adding some liquid support

wAsnk:
As the title says, exposed some viewmodels to liquid templates and added a filter that converts the JToken into the Amount struct. It's needed for liquid template overrides.

### DIFF
--- a/src/Modules/OrchardCore.Commerce/Liquid/AmountConverterFilter.cs
+++ b/src/Modules/OrchardCore.Commerce/Liquid/AmountConverterFilter.cs
@@ -12,7 +12,7 @@ public class AmountConverterFilter : ILiquidFilter
     public ValueTask<FluidValue> ProcessAsync(FluidValue input, FilterArguments arguments, LiquidTemplateContext context)
     {
         if (input.Type == FluidValues.Nil ||
-            input.ToObjectValue() as JToken is not { } newInput)
+            input.ToObjectValue() is not JToken newInput)
         {
             return new ValueTask<FluidValue>(input);
         }
--- a/src/Modules/OrchardCore.Commerce/Liquid/AmountConverterFilter.cs
+++ b/src/Modules/OrchardCore.Commerce/Liquid/AmountConverterFilter.cs
@@ -7,13 +7,17 @@ using System.Threading.Tasks;
 
 namespace OrchardCore.Commerce.Liquid;
 
-public class ToAmountConverterFilter : ILiquidFilter
+public class AmountConverterFilter : ILiquidFilter
 {
     public ValueTask<FluidValue> ProcessAsync(FluidValue input, FilterArguments arguments, LiquidTemplateContext context)
     {
-        if (input.Type == FluidValues.Nil) return new ValueTask<FluidValue>(input);
+        if (input.Type == FluidValues.Nil ||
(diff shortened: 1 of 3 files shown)


wAsnk:
@DAud-IcI or @DemeSzabolcs could you review this?

src/Modules/OrchardCore.Commerce/Startup.cs
```diff
@@ -143,6 +147,18 @@ public override void ConfigureServices(IServiceCollection services)
...
+        services.Configure<TemplateOptions>(option =>
+        {
+            option.MemberAccessStrategy.Register<ShoppingCartViewModel>();
+            option.MemberAccessStrategy.Register<ShoppingCartCellViewModel>();
+            option.MemberAccessStrategy.Register<ShoppingCartLineViewModel>();
+            option.MemberAccessStrategy.Register<Amount, string>((obj, _) => obj.ToString());
+            option.MemberAccessStrategy.Register<Amount, decimal>((obj, _) => obj.Value);
+        })
+        // Liquid filter to convert JToken value to Amount struct in liquid.
+        .AddLiquidFilter<ToAmountConverterFilter>("toAmount");
```
DAud-IcI, line 161: Liquid filters are supposed to be snake_case ([for example](https://docs.orchardcore.net/en/dev/docs/reference/modules/Liquid/#absolute_url)). I think for conversion filters you are also just supposed to name it by the type, so `amount`.
Also please indent this.

src/Modules/OrchardCore.Commerce/Liquid/ToAmountConverterFilter.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace OrchardCore.Commerce.Liquid;
+
+public class ToAmountConverterFilter : ILiquidFilter
+{
+    public ValueTask<FluidValue> ProcessAsync(FluidValue input, FilterArguments arguments, LiquidTemplateContext context)
+    {
+        if (input.Type == FluidValues.Nil) return new ValueTask<FluidValue>(input);
+
+        var newObject = (input.ToObjectValue() as JToken).ToObject<Amount>();
```
DAud-IcI, line 16: `as` can return `null` if the type doesn't match, so this filter will throw NRE if used on the wrong input.

DAud-IcI (changes requested):

src/Modules/OrchardCore.Commerce/Liquid/AmountConverterFilter.cs
```diff
@@ -7,13 +7,17 @@
...
 namespace OrchardCore.Commerce.Liquid;
 
-public class ToAmountConverterFilter : ILiquidFilter
+public class AmountConverterFilter : ILiquidFilter
 {
     public ValueTask<FluidValue> ProcessAsync(FluidValue input, FilterArguments arguments, LiquidTemplateContext context)
     {
-        if (input.Type == FluidValues.Nil) return new ValueTask<FluidValue>(input);
+        if (input.Type == FluidValues.Nil ||
+            input.ToObjectValue() as JToken is not { } newInput)
```
DAud-IcI, line 15:
```suggestion
            input.ToObjectValue() is not JToken newInput)
```

PR #224, opened 2022-10-27
OCC-108: Replace en-EU culture with special EUR handling

### LINKED ISSUE #223: en-EU culture is not available on every OS
### Describe the bug

en-EU culture is not available on every OS, thus OCC causes exceptions once enabled.
On some OS the `new CultureInfo("en-EU")` code creates a culture with the EnglishName property set to `English (European Union)` , but on Azure e.g. the EnglishName will be `Unknown Locale (en-EU)` and because of that difference ArgumentOutOfRangeException will be thrown at Currency.cs:40.

When the condition is modified before the exception throw to the below code, then it will run fine, but you can't select EUR on the Currency settings page.
```
if ((culture.EnglishName.StartsWith("Unknown Locale", StringComparison.Ordinal) && culture.Name != "en-EU") ||
            culture.EnglishName.StartsWith("Invariant Language", StringComparison.Ordinal))
```

### To Reproduce

1. Run the web app on a different OS than Windows 10 or above. E.g. Azure app service
2. Enable OCC
3. See exceptions on OCC settings pages and other settings pages.

### Expected behavior

EUR can be selected on the currency settings page and no exceptions are thrown because of the non-existant en-EU locale.

DAud-IcI:
OCC-108
Fixes #223

### DIFF
--- a/src/Libraries/OrchardCore.Commerce.MoneyDataType/Currency.extra.cs
+++ b/src/Libraries/OrchardCore.Commerce.MoneyDataType/Currency.extra.cs
@@ -10,6 +10,12 @@ public readonly partial struct Currency
 {
     private static readonly ICurrencyProvider _defaultProvider = new CurrencyProvider();
 
+    public static ICurrency UnspecifiedCurrency { get; } = new Currency("Unspecified", "Unspecified", "---", "---");
+
+    // This is a special case (rendered with specific formatting with invariant culture) due to the currency's
+    // international nature.
+    public static Currency Euro { get; } = new("European Union", "European Union", "€", "EUR");
+
     public static ICurrency UnitedArabEmiratesDirham => _defaultProvider.GetCurrency("AED");
     public static ICurrency AfghanAfghani => _defaultProvider.GetCurrency("AFN");
     public static ICurrency AlbanianLek => _defaultProvider.GetCurrency("ALL");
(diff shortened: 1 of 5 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore.Commerce?pullRequest=224) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore.Commerce?pullRequest=224) before we can accept your contribution.<br/><br/>:x: DAud-IcI [sign now](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore.Commerce?pullRequest=224)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/OrchardCMS/OrchardCore.Commerce?pullRequest=224) it.</sub>

src/Libraries/OrchardCore.Commerce.MoneyDataType/Currency.extra.cs
```diff
@@ -161,6 +184,9 @@ public static ICurrency FromRegion(RegionInfo region, IEnumerable<ICurrencyProvi
 
     public static ICurrency FromCulture(CultureInfo culture, IEnumerable<ICurrencyProvider> providers = null)
     {
+        ArgumentNullException.ThrowIfNull(culture);
+        if (culture.TryGetRegionInfo().ISOCurrencySymbol == "EUR") return Euro;
```
wAsnk, line 188:
```suggestion
        if (culture.TryGetRegionInfo()?.ISOCurrencySymbol == "EUR") return Euro;
```
Possible NRE.

wAsnk (changes requested):

src/Libraries/OrchardCore.Commerce.MoneyDataType/Currency.extra.cs
DAud-IcI: I assumed that's the same case as the culture being null, but you are right. But then it's best to refactor the whole thing to handle null culture or invalid culture that doesn't return a region, and return UnspecifiedCurrency in both cases. Then this method never returns null.

DemeSzabolcs (approved):

PR #226, opened 2022-11-04
Update to OC 1.5.0

Skrypt:
Same as title 😄

### DIFF
--- a/docs/resources/libraries/README.md
+++ b/docs/resources/libraries/README.md
@@ -4,7 +4,7 @@ The below table lists the different .NET libraries used in Orchard Core Commerce
 
 | Library | Usage | Version | License |
 |--- | --- | --- | --- |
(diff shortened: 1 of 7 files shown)


0liver:
That was fast!
🏎

DAud-IcI:
You've got an unneeded `using` according to the static code analysis warning.

Also https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/issues/221 is a blocker to this as Lombiq.HelpfulLibraries and others need to be upgraded to OC 1.5.0 first to avoid version conflicts.

Skrypt:
I have it updated locally but I need to understand how you guys update the git submodules. Create a new branch in each of them? Using this extensions project makes this a little more complex to manage. I would have thought that I could have used an Orchard Core Commerce module without requiring to use all of these extension modules? 🤔

PR #227, opened 2022-11-15
OCC-31: Add product discounts

### LINKED ISSUE #136: Add product discounts (OCC-31)
Note: see #9 for general discussions about various types of promotions. This issue is a subset of that.

### Concept

Let's start with simple discounts associated with a specific product. The discount would be described with a separate `DiscountPart`. Abstraction should be similar to how prices are done in OCC. So there should be an `IPromotionProvider` that describes the service that interacts with the `ShoppingCartItem` (e.g. applies the discount from its content item's `DiscountPart`) and an `IPromotionService` that resolves and applies the promotion providers.

### DiscountPart Fields or Properties

- Percentage (`int`/`NumericField`) a percentage discount to be applied to the product price.
  How should rounding be handled? (for reference [here](https://github.com/bleroy/Nwazet.Commerce/blob/master/Services/Discount.cs#L93) is what Nwazet did)
- Amount (`Amount`) a flat number to be subtracted from the price (ofc result can't be negative)
- BeginningUtc (`DateTime`/`DateTimeField`) if the current time is lower, the discount is ignored.
- ExpirationUtc (`DateTime`/`DateTimeField`) if the current time is higher, the discount is ignored.
- Minimum (`int`/`NumericField`) if the quantity is lower, the discount is ignored.
- Maximum (`int`/`NumericField`) if the quantity is higher, the discount is ignored (0 should be treated as infinite).

You shouldn't be able to have both Percentage and Amount (use handler).

### Services

`IPromotionService` should have `Task<IList<ShoppingCartItem>> AddPromotionsAsync(IList<ShoppingCartItem> items);` ([like price](https://github.com/OrchardCMS/OrchardCore.Commerce/blob/main/src/Modules/OrchardCore.Commerce/Abstractions/IPriceService.cs#L16)). This should sort and select the applicable `IPromotionProvider`s, call the `ApplyingPromotion` event (see below), then apply the resulting promo(s) on the shopping cart item.

`IPromotionProvider` should also have similar members as [its price counterpart](https://github.com/OrchardCMS/OrchardCore.Commerce/blob/main/src/Modules/OrchardCore.Commerce/Abstractions/IPriceProvider.cs), just for promotions. There should be one implementation that uses `DiscountPart`, e.g. `DiscountProvider`.

### Event

- `ApplyingPromotion`: extension point for changing the promo providers' order, or removing some or all of them before they are applied to the shopping cart item.

### Others

- You'll probably have to include a new property in `PrioritizedPrice` to store the old price before the promo was applied. That can be used as a crossed-out price on the product page and cart.
- Update the cart and product views accordingly.
- When is the `IPromotionProvider` applied? I guess it has to be used in `PricePartDisplayDriver` and `PriceVariantsPartDisplayDriver` for display; and in `ShoppingCartController.AddItem` to finalize the offer. In the future we may have other `IPromotionProvider` that works on already added cart items (e.g. coupon codes), but we don't have to prepare for that just yet.

This is just speculation, if you can find a better approach, go for it.

### Tests

Include unit tests with dummy service implementations. Refer to `OrchardCore.Commerce.Tests.PriceTests` for inspiration.

[Jira issue](https://lombiq.atlassian.net/browse/OCC-31)

DemeSzabolcs:
OCC-31

Fixes #136


DemeSzabolcs:
I hit a bit of a roadblock. I created the `PromotionProvider` and `PromotionService` it's working however, I can't apply it to the lines. Technically it works for the total, but not for the individual lines, so the total and the payment amount at the end will be good it just displays the wrong amount in the shopping cart by lines

Here is an example with a 10% discount:
image

It is because of how `LocalTaxProvider` works. Let me explain: So I need to apply the promotion inside [ShoppingCartHelpers.cs](https://github.com/OrchardCMS/OrchardCore.Commerce/blob/main/src/Modules/OrchardCore.Commerce/Services/ShoppingCartHelpers.cs). `_promotionService.AddPromotionsAsync(items)` works with `ShoppingCartItems`. So I added it [here](https://github.com/OrchardCMS/OrchardCore.Commerce/blob/issue/OCC-31/src/Modules/OrchardCore.Commerce/Services/ShoppingCartHelpers.cs#L45). It's all good until [this](https://github.com/OrchardCMS/OrchardCore.Commerce/blob/issue/OCC-31/src/Modules/OrchardCore.Commerce/Services/ShoppingCartHelpers.cs#L79) line. Here the tax is applied. And the problem is that the tax is applied in a way that ignores the original price, and it just overrides the `UnitPrice` with the `GrossPrice`. https://github.com/OrchardCMS/OrchardCore.Commerce/blob/6645257c7dd248e30ee915c6527524bb4d657131/src/Modules/OrchardCore.Commerce/Services/LocalTaxProvider.cs#L35

So I don't know what would be the best solution, since I can't call `_promotionService.AddPromotionsAsync(items)` after the tax is applied, since this function requires `ShoppingCartItem`s not `ShoppingCartLineViewModel`s. I could rewrite the `PromotionService` to use `ShoppingCartLineViewModel`s, though I'm not sure about this approach.

I would appreciate some advice on this! I tag you @DAud-IcI since you created the issue description, and implemented the tax system.

DAud-IcI (changes requested):

src/Modules/OrchardCore.Commerce.Promotion/OrchardCore.Commerce.Promotion.csproj
```diff
@@ -0,0 +1,37 @@
...
+    <TargetFramework>net6.0</TargetFramework>
+    <AddRazorSupportForMvc>true</AddRazorSupportForMvc>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <Title>Orchard Core Commerce - Promotion</Title>
+    <Authors>Bertrand Le Roy</Authors>
+    <Copyright>Copyright © 2018 .NET Foundation</Copyright>
+    <Description>Taxation module for Orchard Core Commerce for discounts and promotions.</Description>
+    <PackageTags>OrchardCore;OrchardCore.Commerce;Commerce;e-Commerce;Tax;Discounts;Promotions</PackageTags>
```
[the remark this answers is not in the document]
DemeSzabolcs: Copy-paste error 😄

src/Modules/OrchardCore.Commerce/Assets/Scripts/discount.js
```diff
@@ -0,0 +1,14 @@
+const grossPrice = document.querySelector('span.tax-part-gross-price-value');
+const defaultPrice = document.querySelector('span.price-part-price-field-value');
```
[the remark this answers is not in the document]
DemeSzabolcs: These two elements are created dynamically (`PriceField.cshtml`), so I can't really put a shared class on them unless I check and match against the two `DisplayName()` (`GrossPrice` and `Price`) in `PriceField.cshtml`.

The other thing is, that if the product has a Tax part, then both the normal price, and the gross price is displayed, and with that approach, it would cross out both of the prices. I think it's enough to cross out the gross price if there is a tax part since the customer will only pay (in this case would have paid) for that.

src/Modules/OrchardCore.Commerce.Promotion/Views/DiscountPart_Edit.cshtml
```diff
@@ -0,0 +1,2 @@
+@using OrchardCore.Commerce.Promotion.ViewModels
+@model DiscountPartViewModel
```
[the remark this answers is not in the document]
DemeSzabolcs: It is needed, otherwise it throws an error. Yes it's meant to be empty.

src/Modules/OrchardCore.Commerce.Promotion/Views/DiscountPart.cshtml
```diff
@@ -0,0 +1,63 @@
...
+{
+    @if (discountPart.DiscountPercentage is { } discountPercentage && discountPercentage.Value > 0)
+    {
+        <div class="field field-type-numericfield field-name-discount-part-discount-percentage">
+            <strong class="field-name-discount-part-discount-percentage-title">@T["Discount Percentage: "]</strong>
+            @T["{0}%", discountPercentage.Value]
+        </div>
+    }
+
+    @if (discountPart.MaximumProducts is { } maximumProducts && maximumProducts.Value > 0)
```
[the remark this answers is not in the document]
DemeSzabolcs: `IsValidAndPositive` is for `Amount` this is just the percentage. If the discount is in `Amount` it is displayed by `PriceField.cshtml`.

src/Modules/OrchardCore.Commerce/Assets/Scripts/discount.js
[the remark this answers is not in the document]
DAud-IcI: Ok, then do Array.from(document.querySelectorAll('.tax-part-gross-price-value, .price-part-price-field-value') and iterate over it.

src/Modules/OrchardCore.Commerce.Promotion/Views/DiscountPart.cshtml
```diff
@@ -0,0 +1,62 @@
...
+
+@model OrchardCore.Commerce.Promotion.ViewModels.DiscountPartViewModel
+
+@{
+    var discountPart = Model.DiscountPart;
+}
+
+@if (discountPart.IsValidAndActive())
+{
+    @if (discountPart.DiscountPercentage is { } discountPercentage && discountPercentage.Value > 0)
```
DAud-IcI, line 13: All of the `DiscountPart` properties are initialized with `new()` so you don't have to check if they are `null` in this file.

src/Modules/OrchardCore.Commerce/Assets/Scripts/discount.js
```diff
@@ -0,0 +1,7 @@
+const prices = Array.from(document.querySelectorAll('.tax-part-gross-price-value, .price-part-price-field-value'));
+
+function strikeOutPrice(price) {
+    price.innerHTML = "<del class=\"text-danger\">" + price.textContent + "</del>";
```
DAud-IcI, line 4: According to ESLint:

    4:23  warning  Strings must use singlequote                         quotes
    4:75  warning  Strings must use singlequote                         quotes

src/Modules/OrchardCore.Commerce/Assets/Scripts/discount.js
```diff
@@ -0,0 +1,7 @@
...
+}
+
+prices.forEach(price => strikeOutPrice(price));
```
DAud-IcI, line 7: According to ESLint:

    7:16  warning  Expected parentheses around arrow function argument  arrow-parens

DAud-IcI (changes requested):

src/Modules/OrchardCore.Commerce.Promotion/Views/DiscountPart.cshtml
DemeSzabolcs: I will keep the `null` check for the DateField values, since the field can't be `null`, but it's value property can.
image

DAud-IcI:
I found a bug:
1. Edit product with currently no discount (e.g. https://localhost:5001/Admin/Contents/ContentItems/testproduct000/Edit?returnUrl=%2FAdmin%2FContents%2FContentItems%3Fadmin%3D-565403957)
2. Click publish without changing anything.
3. You'll get this exception:
```
System.InvalidOperationException: Can't compare amounts of different currencies (USD and ---).
   at OrchardCore.Commerce.MoneyDataType.Amount.ThrowIfCurrencyDoesntMatch(Amount other, String operation) in E:\aux_c\repositories\orchard-core\OrchardCore.Commerce\src\Libraries\OrchardCore.Commerce.MoneyDataType\Amount.cs:line 84
   at OrchardCore.Commerce.MoneyDataType.Amount.op_LessThan(Amount first, Amount second) in E:\aux_c\repositories\orchard-core\OrchardCore.Commerce\src\Libraries\OrchardCore.Commerce.MoneyDataType\Amount.cs:line 150
   at OrchardCore.Commerce.Handlers.DiscountPartHandler.UpdatedAsync(UpdateContentContext context, DiscountPart instance) in E:\aux_c\repositories\orchard-core\OrchardCore.Commerce\src\Modules\OrchardCore.Commerce\Handlers\DiscountPartHandler.cs:line 49
   at OrchardCore.ContentManagement.Handlers.ContentPartHandler`1.OrchardCore.ContentManagement.Handlers.IContentPartHandler.UpdatedAsync(UpdateContentContext context, ContentPart part)
   at OrchardCore.ContentManagement.Handlers.ContentPartHandlerCoordinator.<>c.<UpdatedAsync>b__27_0(IContentPartHandler handler, UpdateContentContext context, ContentPart part)
   at OrchardCore.Modules.InvokeExtensions.InvokeAsync[TEvents,T1,T2](IEnumerable`1 events, Func`4 dispatch, T1 arg1, T2 arg2, ILogger logger)
```
4. The editor still displays success saying "Your Product has been published."

src/Modules/OrchardCore.Commerce.Promotion/Views/DiscountPart.cshtml
```diff
@@ -0,0 +1,65 @@
...
+
+    @if (discountPart.ExpirationUtc.Value is { } expirationUtc)
+    {
+        <div class="field field-type-datetimefield field-name-discount-part-expiration-utc">
+            <strong class="field-name-discount-part-expiration-utc-title">@T["Discount Expiration Utc:"]</strong>
+            @expirationUtc
+        </div>
+    }
+
+    if (Model.NewPrice.Amount.IsValid)
```
DAud-IcI, line 56: I wonder, is this ever actually false? Also I think if it's false the whole part shouldn't render.

src/Modules/OrchardCore.Commerce.Promotion/Views/DiscountPart.cshtml
```diff
@@ -0,0 +1,65 @@
...
+    {
+        <div class="pb-3 field field-type-pricefield field-name-discount-part-new-price">
+            <strong class="field-name-discount-part-new-title-title">@T["New Price with Discount after Taxes:"]</strong>
+            @Model.NewPrice.Amount
+        </div>
```
DAud-IcI, line 61: Can we move this one to the top?

DAud-IcI (changes requested):

src/Modules/OrchardCore.Commerce.Promotion/Views/DiscountPart.cshtml
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    @if (discountPart.ExpirationUtc.Value is { } expirationUtc)
+    {
+        <div class="field field-type-datetimefield field-name-discount-part-expiration-utc">
+            <strong class="field-name-discount-part-expiration-utc-title">@T["Discount Expiration Utc:"]</strong>
+            @expirationUtc
+        </div>
+    }
```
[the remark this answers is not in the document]
DemeSzabolcs: It seems like the tag helper is not implemented: https://github.com/OrchardCMS/OrchardCore/issues/2965
(It's not working for me.)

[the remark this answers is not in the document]
DemeSzabolcs: But there is a shape, so I can use it like here: https://github.com/OrchardCMS/OrchardCore/blob/ef76b6f2a7faf716bd5f1754a394e9a755e20177/src/OrchardCore.Modules/OrchardCore.Contents/Views/ContentsMeta_SummaryAdmin.cshtml#L10

src/Modules/OrchardCore.Commerce.Promotion/Views/DiscountPart.cshtml
DemeSzabolcs: It's just a failsafe, but yeah the part shouldn't render if it's false.

src/Modules/OrchardCore.Commerce/Services/DiscountProvider.cs
```diff
@@ -0,0 +1,95 @@
...
+    private static bool IsApplicablePerItem(PromotionAndTaxProviderContextLineItem item)
+    {
+        var discountParts = item.Content
+            .ContentItem
+            .OfType<DiscountPart>();
+
+        return discountParts.Any(discountPart => IsApplicablePerDiscountPart(discountPart, item.Quantity));
+    }
+
+    // Incase we have multiple discount parts on one product.
```
DAud-IcI, line 55:
```suggestion
    // In case we have multiple discount parts on one product.
```

DAud-IcI:
I was sooo ready to approve this, but unfortunately I've found another problem:

1. Buy a product without discount.
2. Go to the product's editor.
3. Add a discount percentage or amount.
4. Publish.
5. View page for the already existing order: the subtotal and total has changed.

Expectation: promotions should not affect existing offers retroactively.

You probably have to store the current ContentItem version in the `OrderPart.LineItems`.

src/Modules/OrchardCore.Commerce/Services/DiscountProvider.cs
```diff
@@ -0,0 +1,95 @@
...
+    private static bool IsApplicablePerDiscountPart(DiscountPart discountPart, int itemQuantity)
+    {
+        var discountMaximumProducts = discountPart.MaximumProducts.Value;
+
+        return discountPart.IsValidAndActive() &&
+               !(discountPart.BeginningUtc.Value > DateTime.UtcNow ||
+               discountPart.ExpirationUtc.Value < DateTime.UtcNow ||
```
DAud-IcI, line 62: Using `DateTime.UtcNow` is not okay for when you review existing orders, because it will remove the discount when it's no long in period. You should add a PurchaseDate property to the `PromotionAndTaxProviderContext` record. You can set it to `DateTime.UtcNow` during checkout, but it should be the Order's ContentItem.PublishDate when it's coming from `OrderPartDisplayDriver.PopulateViewModelAsync`.

DAud-IcI (changes requested):

DemeSzabolcs:
> I was sooo ready to approve this, but unfortunately I've found another problem:
> 
> 1. Buy a product without discount.
> 2. Go to the product's editor.
> 3. Add a discount percentage or amount.
> 4. Publish.
> 5. View page for the already existing order: the subtotal and total has changed.
> 
> Expectation: promotions should not affect existing offers retroactively.
> 
> You probably have to store the current ContentItem version in the `OrderPart.LineItems` and use that instead of looking up the product by SKU in `IOrderLineItemService.CreateOrderLineItemViewModelsAndTotalAsync()`.

FYI: This was effecting taxation too in the same way, but this should solve that too.

DAud-IcI:
> FYI: This was effecting taxation too in the same way, but this should solve that too.

Excellent, thanks for catching that!

PR #228, opened 2022-11-18
OCC-101: New fields of OrderPart are not shown

### LINKED ISSUE #211: New fields of OrderPart are not shown (OCC-101)
### Describe the bug

There was a change on how OCC shows the checkout page and that caused a bug that prevents showing new fields on the checkout page. Currently if you add a new field to the OrderPart it is not shown.

### To Reproduce

1. Add a new field to OrderPart through the UI
2. Add a product to cart then checkout
3. Newly added field is missing, due to the custom shape.

### Expected behavior

When you add a new field to the OrderPart it should show up on the checkout page, on the successful payment page, on the order content item editor and view. So basically where the OrderPart is used.

[Jira issue](https://lombiq.atlassian.net/browse/OCC-101)

DAud-IcI:
OCC-101

Fixes #211


src/Modules/OrchardCore.Commerce/Services/FieldsOnlyDisplayManager.cs
```diff
@@ -0,0 +1,107 @@
...
+        ContentItem contentItem,
+        string displayType = CommonContentDisplayTypes.Detail)
+    {
+        var fieldShapeTypes = GetFieldShapeTypes(contentItem, displayType);
+
+        var shapes = new List<IShape>();
+        foreach (var shapeType in fieldShapeTypes)
+        {
+            shapes.Add(await _shapeFactory.CreateAsync(shapeType));
+        }
```
DemeSzabolcs, line 67: Isn't something like this would simpler? 
```suggestion
        var shapes = (await Task.WhenAll(fieldShapeTypes.Select(async shapeType =>
            await _shapeFactory.CreateAsync(shapeType)))).ToList();
```

DemeSzabolcs (changes requested):

DemeSzabolcs (approved):

Issue #149, opened 2022-08-15
Separate Stripe payment into a feature so it can be disabled if other payment provider is desired. (OCC-63)


sarahelsaig:
This won't be urgent until we have other payment providers. If someone needs it soon because you already have your own payment implementation, please comment on this issue.

- Refactor the Stripe payment processor logic into a separate project, e.g. OrchardCore.Commerce.PaymentProcessors.Stripe.
- Create the necessary abstraction to make it provider-based with Stripe being just one provider.
  - Replace the `~/cart`'s <kbd>Checkout</kbd> button with a <kbd>Checkout with {providerName}</kbd> button so it's locked down in the order form. Then only display the correct form for that processor.
  - Use the `IPaymentProvider` and `IPayment` interfaces.
- Add a dummy payment processor provider.
  - This should be a separate feature.
  - It should be enabled in the recipe but (if possible) force-disabled in production.
  - Add a sample `Order` content item to the recipe, whose payment info references this provider.
- Include which payment provider was used in the `OrderPart`.
  - Make it read-only if the Order is already created and this has been set.

Also create a unit test of the basic purchase flow using the dummy payment processor. 
- Buy different amounts of the two sample products.
- Change the amount of one in the cart.
- Fill the checkout form.
- Go to the newly created Order item.
- Verify that the order has the correct products an amounts, and that the correct payment provider name is displayed.

If #155 is resolved, then also set the addresses, and verify that they are valid as well. 

[Jira issue](https://lombiq.atlassian.net/browse/OCC-63)

hishamco:
I'm working on a project that uses a sort of eCommerce, I agree to separate Stripe payment but not as a feature. It could be as the following:

- OCC.Payment.Abstractions
- OCC.Payment.Core (Stripe Payment Implementation)
- OCC.Payment (Module)

This way it will be much flexible and extensible

DAud-IcI:
It would be important to be able to disable unused payment implementations. For example if we have Stripe for production and a dummy payment processor for UI testing. I suggested using features because that seems like a natural fit for this purpose in OC. Do you have something else in mind?

hishamco:
`OCC.Payment` module will have the default implementation, so it could be enabled and disabled. As a user, I can create `OCC.Payments.PayPal` enable it and disable the other provider

In another scenario, some websites might support multiple gateway providers

hishamco:
[...]

PR #236, opened 2022-12-11
OCC-115: Migrate to payment element from card element

wAsnk:
OCC-115
Followed docs [https://stripe.com/docs/payments/payment-element/migration](https://stripe.com/docs/payments/payment-element/migration). Had to made some changed in the workflow of how the payment intent is created and handled also did some refactoring in the logic. 
Automatically listing payment methods in the payment intent element now.
Using the recommended webhook events instead of depending on the return url.

Supported payment methods by this PR:
- Card
- iDeal
- Google pay
- giropay
- Bancontact
- EPS
- Alipay
- Przelewy24
- Link
- Apple Pay most possibly is working (uses the same procedure as Google pay), but needs extra steps on the Stripe dashboard and a file upload on a live site exposed to the internet

Stripe webhook local test:
1. [Install the Stripe CLI](https://stripe.com/docs/stripe-cli) and log in to authenticate your account.
2. Set listening endpoint `stripe listen --forward-to https://localhost:[port]/stripe-webhook`
3. Checkout and pay.


DAud-IcI (changes requested):

src/Modules/OrchardCore.Commerce/Assets/Scripts/stripe-payment-form.js
```diff
@@ -1,34 +1,43 @@
...
-        style: {
-            base: {
-                fontWeight: '500',
-                fontSize: '20px',
-            },
-        },
+    const checkboxTypeName = 'checkbox';
+    const stripeElements = stripe.elements({
+        clientSecret: clientSecret,
+    });
```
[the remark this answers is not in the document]
wAsnk: Oh shoot I didn't check this...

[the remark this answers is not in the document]
wAsnk: Weirdly this is done by Rider by default on document formatting.

src/Modules/OrchardCore.Commerce/Controllers/PaymentController.cs
```diff
@@ -254,27 +275,40 @@ public async Task<IActionResult> Pay(string paymentMethodId, string paymentInten
...
+        if (paymentIntent.Status == PaymentIntentStatuses.RequiresAction &&
             paymentIntent.NextAction.Type == "use_stripe_sdk")
         {
             // Tell the client to handle the action.
-            return Json(new
-            {
-                requires_action = true,
-                payment_intent_client_secret = paymentIntent.ClientSecret,
-            });
+            return Json(new { requires_action = true, payment_intent_client_secret = paymentIntent.ClientSecret, });
```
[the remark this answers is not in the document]
wAsnk: No idea, it was like this before. But good catch!

src/Modules/OrchardCore.Commerce/Assets/Scripts/stripe-payment-form.js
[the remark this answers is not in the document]
DAud-IcI: I don't think I have ever wanted to do that on JS but if you are going to use it, try to [import the settings from our ESLint config](https://stackoverflow.com/a/60526644/14748624). Nothing about JS code style is standardized so you can't rely on an IDE's defaults to match with the linter's.

src/Modules/OrchardCore.Commerce/Abstractions/IStripePaymentService.cs
```diff
@@ -8,13 +8,19 @@ namespace OrchardCore.Commerce.Abstractions;
 /// <summary>
 /// When implemented handles the payment and creates an order.
 /// </summary>
-public interface ICardPaymentService
+public interface IStripePaymentService
```
Skrypt, line 11: This will break a lot of things. It was meant to be a generic Interface for any payment provider.

wAsnk: This interface was already for Stripe only before this PR.

Skrypt: In the OrchardCore.Commerce module you mean. Of course, there is only Stripe supported out of the box but nothing prevents me from using this Interface and creating my own. Now, renaming it means that I need to use the IStripePaymentService to create a IPaypalPaymentService which doesn't make any sense.

wAsnk: Yes, you are right. But my point is that it wasn't supported even before this PR. Abstracting payment services and supporting other possible payment providers are not part of this PR and also not part of OCC yet.

Skrypt: It has always been a target. That's why we used that Interface name.

Skrypt: This is a Commerce abstraction project ... we talked about this in a meeting a long time ago. I don't see the issue and why we would want to only support Stripe ...

Skrypt: The goal of an abstraction is to be abstract (name included).

Skrypt: Also, about history, they moved the abstractions to different folders than what it was in the beginning. This is why it doesn't show any history for what it was in the beginning. You can't rely on Git history to say whether an Interface is new or not.

.github/workflows/build-and-test.yml
```diff
@@ -18,4 +18,4 @@ jobs:
 
   spelling:
     name: Spelling
-    uses: Lombiq/GitHub-Actions/.github/workflows/spelling.yml@dev
+    uses: Lombiq/GitHub-Actions/.github/workflows/spelling.yml@dev
```
DAud-IcI, line 21: Revert this.

src/Modules/OrchardCore.Commerce/Abstractions/IStripePaymentService.cs
DAud-IcI: This interface was [added for the Stripe support](https://github.com/OrchardCMS/OrchardCore.Commerce/commit/cb23b7c02be1502cbb8decdb55a97f01003f65b3) (it wasn't moved, this is the whole history of the file). Currently it's not possible to have anything else, until proper abstraction is added via #149. (this interface should be moved into a new OrchardCore.Commerce.PaymentProcessors.Stripe module during that issue) It is very Stripe-specific with the whole "payment intent" thing, and a more abstract interface will have to look differently. Because of that I'm certain there are no extensions this change could break.

Skrypt: Ok then. You seem right. I don't see why this interface is required then. It is more confusing than anything and I understand why it is getting renamed now. 

You eventually probably need to make use of the IPaymentProvider and IPayment interfaces instead of ICardPaymentService and PaymentIntent then. Where a payment provider provides multiple ways to do payments.

src/Modules/OrchardCore.Commerce/Controllers/PaymentController.cs
```diff
@@ -245,70 +223,44 @@ public async Task<IActionResult> SuccessPost(string orderId)
...
-        return await GeneratePaymentResponseAsync(paymentIntent);
+        // Set back to default, because a new payment intent should be created on the next checkout.
+        _paymentIntentPersistence.Store(paymentIntentId: string.Empty);
     }
 
-    private async Task<IActionResult> GeneratePaymentResponseAsync(PaymentIntent paymentIntent)
+    [AllowAnonymous]
+    [HttpGet("checkout/middleware")]
+    // Sadly that is how it is given.
+#pragma warning disable CA1707
```
DAud-IcI, line 239: Include the warning message too, like it is with the suppression on line 58. But if you don't mind, please turn both into `[SuppressMessage]` attributes because that's cleaner for warnings in the method header.

Also you could use the Name property of `[FromQuery]` ([see](https://stackoverflow.com/a/38307063/14748624)) and keep the variable name camelCase.

src/Modules/OrchardCore.Commerce/Controllers/PaymentController.cs
```diff
@@ -245,70 +223,44 @@ public async Task<IActionResult> SuccessPost(string orderId)
...
-            var order = await _stripePaymentService.CreateOrderFromShoppingCartAsync(paymentIntent);
+        var fetchedPaymentIntent = await _stripePaymentService.GetPaymentIntentAsync(payment_intent);
+        var orderId = (await _stripePaymentService.GetOrderPaymentByPaymentIntentIdAsync(payment_intent))?.OrderId;
 
-            return Json(new { Success = true, OrderContentItemId = order.ContentItemId });
-        }
+        var order = await _contentManager.GetAsync(orderId);
+        var status = order?.As<OrderPart>()?.Status?.Text;
+        var finished = fetchedPaymentIntent.Status == PaymentIntentStatuses.Succeeded &&
+                       !string.IsNullOrEmpty(orderId) &&
```
DAud-IcI, line 249: This should be quicker if you have to fetch `order` anyway.
```suggestion
                       order != null &&
```

src/Modules/OrchardCore.Commerce/Controllers/WebhookController.cs
```diff
@@ -0,0 +1,83 @@
...
+    public async Task<IActionResult> Index()
+    {
+        using var streamReader = new StreamReader(HttpContext.Request.Body);
+        var json = await streamReader.ReadToEndAsync();
+        try
+        {
+            // If the webhook signing key is empty, default to the test key.
+            var stripeApiSettings = (await _siteService.GetSiteSettingsAsync()).As<StripeApiSettings>();
+            var webhookSigningKey = stripeApiSettings.DecryptWebhookSigningSecret(_dataProtectionProvider, _logger);
+            webhookSigningKey = string.IsNullOrEmpty(webhookSigningKey) ? LocalEndPointSecret : webhookSigningKey;
```
DAud-IcI, line 52: You can simplify this by using `.OrIfEmpty(LocalEndPointSecret)` .

src/Modules/OrchardCore.Commerce/Controllers/WebhookController.cs
```diff
@@ -0,0 +1,83 @@
...
+                var charge = stripeEvent.Data.Object as Charge;
+                if (charge?.PaymentIntentId is not { } paymentIntentId)
+                {
+                    return BadRequest();
+                }
+
+                var paymentIntent = await _stripePaymentService.GetPaymentIntentAsync(paymentIntentId);
+                await _stripePaymentService.UpdateOrderToOrderedAsync(paymentIntent);
+            }
+            else if (stripeEvent.Type == Stripe.Events.PaymentIntentPaymentFailed)
```
DAud-IcI, line 70: Don't you need to handle an `else` case after this? What happens if the event type is neither?

src/Modules/OrchardCore.Commerce/Migrations/OrderMigrations.cs
```diff
@@ -202,8 +197,18 @@ public int UpdateFrom3()
...
+            );
 
         return 4;
     }
+
+    public int UpdateFrom4()
+    {
+        SchemaBuilder
+            .CreateMapIndexTable<OrderPaymentIndex>(table => table
+                .Column<string>(nameof(OrderPaymentIndex.OrderId), column => column.WithLength(26))
```
DAud-IcI, line 209:
```suggestion
                .Column<string>(nameof(OrderPaymentIndex.OrderId), column => column.WithCommonUniqueIdLength())
```

src/Modules/OrchardCore.Commerce/Services/StripePaymentService.cs
```diff
@@ -128,10 +183,34 @@ public async Task<ContentItem> CreateOrderFromShoppingCartAsync(PaymentIntent pa
...
-        var order = await _contentManager.NewAsync("Order");
-        var orderId = Guid.NewGuid();
+        var orderId = (await GetOrderPaymentByPaymentIntentIdAsync(paymentIntent.Id))?.OrderId;
 
-        order.DisplayText = T["Order {0}", orderId];
+        ContentItem order;
+        string guidId;
+        if (string.IsNullOrEmpty(orderId))
+        {
+            order = await _contentManager.NewAsync("Order");
```
DAud-IcI, line 192:
```suggestion
            order = await _contentManager.NewAsync(Constants.ContentTypes.Order);
```

src/Modules/OrchardCore.Commerce/Views/Payment/PaymentConfirmationMiddleware.cshtml
```diff
@@ -0,0 +1,23 @@
...
+        await new Promise(resolve => {
+            return setTimeout(resolve, milliseconds)
+        });
+    };
+    const reload = async () => {
+        await sleep(2000);
+        location.reload();
+    }
+
+    reload();
```
DAud-IcI, line 22: Do you really need a whole Promise wrapper in an inline script?
Whats wrong with just `setTimeout(() => location.reload(), 2000)`?

DAud-IcI (changes requested):

src/Modules/OrchardCore.Commerce/Controllers/WebhookController.cs
wAsnk: It is recommended to return Ok in other scenarios, which we don't want to handle:
[https://stripe.com/docs/webhooks#built-in-retries](https://stripe.com/docs/webhooks#built-in-retries)

wAsnk:
> I set up the local stripe CLI but it keeps looping for me on the "Confirming payment." page. The payment intent status is "succeeded" but the order status is still "pending".
> 
> image
> 
> image

Could you write me down the repro steps because it works on a fresh clone for me. What I do:

0. Enable OCC features.
1. Set default currencies to EUR. 
2. Set API pubkey and secret key to the publicly available keys (You don't need to set the webhook key for local testing BCS that's the default)
3. Run OC Product recipes 
4. Set TestProduct currency to EUR
5. Run `stripe listen --forward-to https://localhost:5001/stripe-webhook`. (You don't even need to login)
6. When you add the product to the cart then you checkout you should already see these:
image
When you select a bank redirect method (e.g. iDeal) you should get this:
image
When the payment succeeds you should get these two:
image

DAud-IcI:
1. Nothing shows up in my console when I click checkout:
image

2. Is switching to EUR actually necessary? If yes, please explain why in the readme. Also I please update the Stripe setup instructions there.

3. I did a new setup and then after the first try I keep getting this error. It seems to generate the same payment intent.
image

src/Modules/OrchardCore.Commerce/Controllers/WebhookController.cs
```diff
@@ -0,0 +1,84 @@
...
+
+[Route("stripe-webhook")]
+[ApiController]
+[Authorize(AuthenticationSchemes = "Api"), IgnoreAntiforgeryToken, AllowAnonymous]
+public class WebhookController : Controller
+{
+    // This is a default test Stripe CLI webhook secret for testing your endpoint locally. It can be shown, as this is
+    // the same always for testing in Stripe.
+    private const string LocalEndPointSecret =
+        "whsec_453d1046fc31377b7a93e82b839c9e6e065d7117b6e02422e55eac99da087463";
```
DAud-IcI, line 24: Where does this secret come from? I can't find it anywhere in the documentations and when I run `stripe` it shows a different secret. Please include a link

src/Modules/OrchardCore.Commerce/Controllers/PaymentController.cs
```diff
@@ -174,24 +189,8 @@ public async Task<IActionResult> Success(string orderId)
         return View(order);
     }
 
-    [Route("success/{orderId}")]
-    [HttpPost]
-    [ValidateAntiForgeryToken]
-    public async Task<IActionResult> SuccessPost(string orderId)
+    private async Task FinalModificationAsync(ContentItem order)
```
DAud-IcI, line 192: This method contains 
```csharp
        order.Alter<OrderPart>(part => part.Status =
            new TextField { ContentItem = order, Text = OrderStatuses.Ordered.HtmlClassify() });
```
but it's only called by `PaymentConfirmationMiddleware` if the order status is already "ordered". So this alter is not necessary.

DAud-IcI (changes requested):

src/Modules/OrchardCore.Commerce/Controllers/PaymentController.cs
```diff
@@ -224,57 +223,41 @@ public async Task<IActionResult> SuccessPost(string orderId)
...
-            // Create the order content item.
-            var order = await _cardPaymentService.CreateOrderFromShoppingCartAsync(paymentIntent);
-
-            return Json(new { Success = true, OrderContentItemId = order.ContentItemId });
+            return RedirectToAction(nameof(Index));
         }
 
-        // Invalid status.
-        return StatusCode(StatusCodes.Status500InternalServerError, new { error = T["Invalid PaymentIntent status"].Value });
+        return View();
```
DAud-IcI, line 261: I think you should only `return View();` if the payment status is `PaymentIntentStatuses.Processing`. You should check that right after the first line of the method to avoid pointlessly querying the order.

Looks like every status besides that is an error if you have reached this point. Including when the payment status is "succeeded" but the `finished` variable above is not true. In this case if you keep looping this site every 2 seconds, then you waste the web server on an impossible request loop, and also waste stripe API requests (I assume there is some quota in production).
image
You should redirect back to the checkout page and display an error notification.

src/Modules/OrchardCore.Commerce/Controllers/WebhookController.cs
wAsnk: Okay, so it might have been a generated local test key for me as well. It's coming from here:
[https://dashboard.stripe.com/test/webhooks/create?endpoint_location=local](https://dashboard.stripe.com/test/webhooks/create?endpoint_location=local)

src/Modules/OrchardCore.Commerce/Controllers/PaymentController.cs
wAsnk: Implemented a shortcut here and added error notifier. Although the webhook should work in case the user don't get redirected for any reason, but the payment still goes through.

DAud-IcI:
Since your latest changes, now checkout doesn't get stuck any more. 👍

PR #238, opened 2022-12-20
OCC-37: Inventory management

### LINKED ISSUE #134: Inventory management (OCC-37)
### Concept

Content items which have `ProductPart` may also have an `InventoryPart`. If a type has `InventoryPart`, then additional conditions apply before the product can be added to the cart.

### Fields or Properties

Should this use OrchardCore.ContentFields? It would be nicer, I think.

- **Inventory** (int/NumericField) The number of items in stock.
- **AllowsBackOrder** (bool/BooleanField) If true, ordering is allowed even if Inventory is less or equal to zero.
- **OutOfStockMessage** (string/MarkdownField/HtmlField) Enables a specific message for an out of stock product (can be used to give an ETA).
- **IgnoreInventory** (bool/BooleanField) If true, inventory is ignored same as not having an `InventoryPart` at all. This is useful for digital products for example.
- **MinimumOrderQuantity** (int/NumericField) The minimum number of this item one can order; ignored if zero or negative.
- **MaximumOrderQuantity** (int/NumericField) The maximum number of this item one can order (per day? ever?); ignored if zero or negative.

I have changed the ConsiderInventory from the original proposal into IgnoreInventory so it can be default false.

### Events

Should have an `IInventoryEvents` interface with at least
- AddingToCart: before item is added to cart; add custom logic e.g. authentication check and cancel adding to cart if needed.
- OrderCheckedOut: after payment; should contain a default handler for updating the inventory, this is where we'll trigger workflows too.

### Services and Providers

- `IProductInventoryService` & `ProductInventoryService` (Product prefix so the implementation doesn't start with the letter I.)
  - `Task<int> QueryInventoryAsync(string sku)`: return the current inventory count. (also suggest a `Task<bool> IsAvailableAsync(string sku)` as an extension method)
  - `Task UpdateInventoryAsync(string sku, int difference, bool reset = false)`: Normally you want to change the inventory by decrementing with sales amount or incrementing with shipment amount. Reset sets the value to zero first. This should be locking/atomic to prevent problems with simultaneous orders.
- `IProductInventoryProvider` & `LocalInventoryProvider`:
  - Include the actual implementations of the two methods above.
  - `Order` property.
  - `Task<bool> IsApplicableAsync(string sku)`: check if it product can use this provider. For `LocalInventoryProvider` this means the presence of the `InventoryPart`.

### Workflows

They are great, but it should be part of a followup issue to tighten the scope.

### Project Structure

The events, interfaces and services (excluding `LocalInventoryProvider`) should go to OrchardCore.Commerce.
`LocalInventoryProvider` and `InventoryPart` should go it a new module, e.g. OrchardCore.Comerce.Inventory.Local. Future ERP-based implementations would go to OrchardCore.Comerce.Inventory.{ProductName} projects accordingly.

porgabi:
OCC-37

Fixes #134


src/Modules/OrchardCore.Commerce/Extensions/ProductInventoryProviderExtensions.cs
```diff
@@ -0,0 +1,9 @@
+using System.Threading.Tasks;
+
+namespace OrchardCore.Commerce.Abstractions;
+
+public static class ProductInventoryProviderExtensions
+{
+    public static async Task<bool> IsAvailableAsync(this IProductInventoryProvider provider, string sku) =>
```
DAud-IcI, line 7: You should never work with a single provider. Always use `IEnumerable<IProductInventoryProvider>`. You can get the currently applicable provider using `var provider = await providers.GetFirstApplicableProviderAsync(model)`.

src/Modules/OrchardCore.Commerce/Services/CardPaymentService.cs
```diff
@@ -45,7 +47,9 @@ public class CardPaymentService : ICardPaymentService
         IDataProtectionProvider dataProtectionProvider,
         ILogger<CardPaymentService> logger,
         IStringLocalizer<CardPaymentService> stringLocalizer,
-        ISession session)
+        ISession session,
+        IProductInventoryProvider productInventoryProvider,
+        IProductService productService)
```
DAud-IcI, line 52: Don't inject `IProductInventoryProvider` (it should be `IEnumerable<IProductInventoryProvider>` anyway). You should use `_productService.UpdateAsync(item)` to update the inventory without having to know which provider is available.

src/Modules/OrchardCore.Commerce/Services/ProductInventoryProvider.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        return model.All(item =>
+            skuProducts.TryGetValue(item.ProductSku, out var productPart) &&
+            productPart.ContentItem.Has<InventoryPart>());
+    }
+
+    public async Task<int> QueryInventoryAsync(string sku)
+    {
+        var inventoryPart = (await _productService.GetProductAsync(sku))?.As<InventoryPart>();
+        return inventoryPart != null ? (int)inventoryPart.Inventory.Value : 0;
```
DAud-IcI, line 41: image
This will crash, because you can't safly cast `decimal?` into `int`. You should do `inventoryPart?.Inventory.Value is { } inventory ? (int)inventory : 0` instead.

src/Modules/OrchardCore.Commerce/Services/ProductInventoryProvider.cs
```diff
@@ -0,0 +1,66 @@
...
+    {
+        foreach (var item in model)
+        {
+            UpdateInventory(await _productService.GetProductAsync(item.ProductSku), item.Quantity);
+        }
+
+        return model;
+    }
+
+    public void UpdateInventory(ProductPart productPart, int difference, bool reset = false)
```
DAud-IcI, line 54: As mentioned above you don't need to expose this so you can turn
```suggestion
    private void UpdateInventory(ProductPart productPart, int difference)
```

src/Modules/OrchardCore.Commerce/Services/ProductInventoryProvider.cs
```diff
@@ -0,0 +1,66 @@
...
+    public int Order => 0;
+
+    public async Task<bool> IsApplicableAsync(IList<ShoppingCartItem> model)
+    {
+        var skuProducts = await _productService.GetSkuProductsAsync(model);
+
+        return model.All(item =>
+            skuProducts.TryGetValue(item.ProductSku, out var productPart) &&
+            productPart.ContentItem.Has<InventoryPart>());
+    }
```
DAud-IcI, line 36: I think we can simply return true here, because this is going to be the fallback inventory manager.

src/Modules/OrchardCore.Commerce/Services/ProductInventoryProvider.cs
```diff
@@ -0,0 +1,66 @@
...
+
+    public void UpdateInventory(ProductPart productPart, int difference, bool reset = false)
+    {
+        var inventoryPart = productPart.As<InventoryPart>();
+        if (inventoryPart == null || inventoryPart.IgnoreInventory.Value) return;
+
+        var newValue = reset ? difference : ((int)inventoryPart.Inventory.Value) + difference;
+
+        inventoryPart.Inventory.Value = newValue > 0 ? newValue : 0;
```
DAud-IcI, line 61: I think if we go into the negative we should throw an exception instead of just setting the inventory to zero, otherwise the product will be oversold.

src/Modules/OrchardCore.Commerce/Services/ProductInventoryProvider.cs
```diff
@@ -0,0 +1,66 @@
...
+
+    public async Task<IList<ShoppingCartItem>> UpdateAsync(IList<ShoppingCartItem> model)
+    {
+        foreach (var item in model)
+        {
+            UpdateInventory(await _productService.GetProductAsync(item.ProductSku), item.Quantity);
+        }
+
+        return model;
+    }
```
DAud-IcI, line 52: This should be made atomic by locking with SemaphoreSlim [as Scott mentioned in the issue](https://github.com/OrchardCMS/OrchardCore.Commerce/issues/134#issuecomment-1322122508).

src/Modules/OrchardCore.Commerce/Services/ProductInventoryService.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        var providers = await _productInventoryProviders
+            .OrderBy(provider => provider.Order)
+            .WhereAsync(provider => provider.IsApplicableAsync(items));
+
+        foreach (var productInventoryProvider in providers)
+        {
+            var result = await productInventoryProvider.UpdateAsync(items);
+            items = result.AsList();
+        }
```
DAud-IcI, line 30: You can just use `_productInventoryProviders.UpdateWithFirstApplicableProviderAsync(items)`.

DAud-IcI (changes requested):
It's definitely on the right track. Thanks for the documentation cleanup!

src/Modules/OrchardCore.Commerce/Abstractions/ISortableUpdaterProvider.cs
```diff
@@ -49,6 +49,26 @@ public static class SortableUpdaterProviderExtensions
...
+    /// <see cref="ISortableUpdaterProvider{TModel}.IsApplicableAsync"/> evaluates to <see langword="true"/>.
+    /// </summary>
+    public static async Task<TProvider> GetFirstApplicableProviderAsync<TModel, TProvider>(
+        this IEnumerable<ISortableUpdaterProvider<TModel>> providers,
+        TModel model)
+        where TProvider : ISortableUpdaterProvider<TModel>
+    {
+        foreach (var provider in providers.OrderBy(provider => provider.Order))
+        {
+            if (provider is TProvider castedProvider && await provider.IsApplicableAsync(model))
```
DAud-IcI, line 63: I'm not sure if this is necessary, you can just use the already existing one like this:
`await providers.GetFirstApplicableProviderAsync(model) as IProductInventoryProvider`
It can't be a problem unless `providers` contains an assortment of different types that inherit from `ISortableUpdaterProvider<T>`, but that shouldn't happen anyway.

src/Modules/OrchardCore.Commerce/Services/LocalInventoryProvider.cs
```diff
@@ -0,0 +1,73 @@
...
+namespace OrchardCore.Commerce.Services;
+
+/// <summary>
+/// A product inventory provider that updates inventories of shopping cart items which have an <c>InventoryPart</c>.
+/// </summary>
+public class LocalInventoryProvider : IProductInventoryProvider
+{
+    private readonly IProductService _productService;
+    private readonly ISession _session;
+    private readonly SemaphoreSlim _lock = new(initialCount: 1);
```
DAud-IcI, line 20: I think the SemaphoreSlim is supposed to be static, since it's supposed to lock across threads.

DAud-IcI (changes requested):

DAud-IcI:
Looks like I'm able to buy more than the maximum quantity. Is it checked anywhere?
image

Also I'm able to buy even when the inventory is zero. I think both of these should be checked in `ShoppingCartController.AddItem` (you could create an event to avoid adding inventory management stuff to this action). Also you should set `ProductPartViewModel.CanBeBought` to false when the inventory is zero.

src/Modules/OrchardCore.Commerce/Services/CardPaymentService.cs
DAud-IcI: Looks like `IProductInventoryService` is not used at all right now. Was this accidentally deleted when you merged from the main branch?

porgabi: As `CardPaymentService` was deleted on dev, this part was moved to `StripePaymentService`. However, instead of the service, I used the providers' `UpdateWithFirstApplicableProviderAsync()` method:
image

Is it wrong to do it this way?

DAud-IcI: You already do that in `ProductInventoryService.UpdateInventoriesAsync()`, so why not call that in `StripePaymentService` as well?

porgabi:
> Looks like I'm able to buy more than the maximum quantity. Is it checked anywhere? image
> 
> Also I'm able to buy even when the inventory is zero. I think both of these should be checked in `ShoppingCartController.AddItem` (you could create an event to avoid adding inventory management stuff to this action). Also you should set `ProductPartViewModel.CanBeBought` to false when the inventory is zero.

I wasn't sure how much of the functionality is to be implemented here, so for starters I only added the inventory decreasement on checkout. I'll add those you mentioned.

DAud-IcI (commented):
Thanks, this is good stuff!
I went ahead and fixed the few problems I have found instead of leaving another round of review. It felt like describing them would take too long compared to just fixing and it's almost year end anyway.

DAud-IcI (approved):

PR #289, opened 2023-04-11
SITE-50: Fixing rounding problems

DemeSzabolcs:
SITE-50

### DIFF
--- a/src/Modules/OrchardCore.Commerce/Services/ShoppingCartHelpers.cs
+++ b/src/Modules/OrchardCore.Commerce/Services/ShoppingCartHelpers.cs
@@ -96,12 +96,10 @@ public class ShoppingCartHelpers : IShoppingCartHelpers
         }
 
         // The values are rounded to avoid storing more precision than what the currency supports.
-        for (var i = 0; i < lines.Count; i++)
+        foreach (var line in lines)
         {
-            var line = lines[i];
-
-            lines[i].LinePrice = line.LinePrice.GetRounded();
-            lines[i].UnitPrice = line.UnitPrice.GetRounded();
+            line.LinePrice = line.LinePrice.GetRounded();
+            line.UnitPrice = line.UnitPrice.GetRounded();
         }
 
         model.Totals.AddRange(totals.Round());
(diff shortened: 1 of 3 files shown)


src/Modules/OrchardCore.Commerce/Services/ShoppingCartHelpers.cs
```diff
@@ -94,6 +95,14 @@ public class ShoppingCartHelpers : IShoppingCartHelpers
             totals = lines.CalculateTotals().ToList();
         }
 
+        for (var i = 0; i < totals.Count; i++)
+        {
+            var total = totals[i];
+
+            // The value is rounded to avoid storing more precision than what the currency supports.
+            totals[i] = new Amount(Math.Round(total.Value, total.Currency.DecimalPlaces), total.Currency);
+        }
```
0liver, line 104: Just like `lines.CalculateTotals()`, move this code into an extension method so it can be called like `totals.RoundTotals()` or something like that.

DAud-IcI: I'd rather you add an `Amount GetRounded()` method to Amount. (I prefer that name because `Round()` could imply mutating the original value.) We want rounding available for individual amounts too. You can add an extension method for `IList<Amount>` if you want.

src/Modules/OrchardCore.Commerce/Services/ShoppingCartHelpers.cs
```diff
@@ -94,6 +95,14 @@ public class ShoppingCartHelpers : IShoppingCartHelpers
             totals = lines.CalculateTotals().ToList();
         }
 
+        for (var i = 0; i < totals.Count; i++)
```
DAud-IcI, line 98: Totals is just the sum value. You should update the `LinePrice` and `UnitPrice` in `lines` as well.

DAud-IcI (changes requested):

src/Modules/OrchardCore.Commerce/Services/ShoppingCartHelpers.cs
```diff
@@ -95,15 +95,16 @@ public class ShoppingCartHelpers : IShoppingCartHelpers
...
+        // The values are rounded to avoid storing more precision than what the currency supports.
+        for (var i = 0; i < lines.Count; i++)
         {
-            var total = totals[i];
+            var line = lines[i];
 
-            // The value is rounded to avoid storing more precision than what the currency supports.
-            totals[i] = new Amount(Math.Round(total.Value, total.Currency.DecimalPlaces), total.Currency);
+            lines[i].LinePrice = line.LinePrice.GetRounded();
+            lines[i].UnitPrice = line.UnitPrice.GetRounded();
```
DAud-IcI, line 104: I don't think you have to index here, you can access the same value via `line.LinePrice`. Also maybe this whole loop can be foreach instead of for?

PR #314, opened 2023-07-07
OCC-169: Add user editor for the non-address checkout fields too

### LINKED ISSUE #313: Add user editor for the non-address checkout fields too (OCC-169)

We now have a couple fields that aren't auto filled:
image

This includes "Phone Number" which is a required field.

We should have another front end editor for them like ~/user/addresses.
Also a matching user editor tab in the admin side.

We could also just add these fields to the already existing saved address editor, but that would make it necessary to change the route from ~/user/addresses and break some backwards compatibility. 

[Jira issue](https://lombiq.atlassian.net/browse/OCC-169)

porgabi:
OCC-169

Fixes #313

### DIFF
--- a/src/Modules/OrchardCore.Commerce/Controllers/UserController.cs
+++ b/src/Modules/OrchardCore.Commerce/Controllers/UserController.cs
@@ -132,7 +132,7 @@ public class UserController : Controller
         var contentItem = user.As<ContentItem>(contentType);
 
         return string.IsNullOrEmpty(contentItem?.ContentType)
-            ? await _contentManager.NewAsync(UserAddresses)
+            ? await _contentManager.NewAsync(contentType)
             : contentItem;
     }
 }
--- a/src/Modules/OrchardCore.Commerce/Controllers/UserController.cs
+++ b/src/Modules/OrchardCore.Commerce/Controllers/UserController.cs
@@ -51,7 +51,7 @@ public class UserController : Controller
         if (User.Identity?.IsAuthenticated != true) return LocalRedirect("~/Login?ReturnUrl=~/user/addresses");
         if (await _userManager.GetUserAsync(User) is not User user) return NotFound();
 
-        var userAddresses = await GetUserAddressesAsync(user);
+        var userAddresses = await GetUserContentItemAsync(user, UserAddresses);
         var editor = await _contentItemDisplayManager.BuildEditorAsync(
             userAddresses,
             _updateModelAccessor.ModelUpdater,
@@ -68,7 +68,7 @@ public class UserController : Controller
         if (User.Identity?.IsAuthenticated != true) return LocalRedirect("~/Login?ReturnUrl=~/user/addresses");
         if (await _userManager.GetUserAsync(User) is not User user) return NotFound();
 
-        var userAddresses = await GetUserAddressesAsync(user);
+        var userAddresses = await GetUserContentItemAsync(user, UserAddresses);
         await _contentItemDisplayManager.UpdateEditorAsync(userAddresses, _updateModelAccessor.ModelUpdater, isNew: false);
 
         if (_updateModelAccessor.ModelUpdater.ModelState.IsValid)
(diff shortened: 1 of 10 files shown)


src/Modules/OrchardCore.Commerce/Controllers/UserController.cs
```diff
@@ -86,12 +86,53 @@ public async Task<IActionResult> AddressesPost()
...
+        else
+        {
+            var errors = _updateModelAccessor.ModelUpdater.GetModelErrorMessages();
+            foreach (var error in errors.WhereNot(string.IsNullOrEmpty)) await _notifier.ErrorAsync(H[error]);
+        }
+
+        return RedirectToAction(nameof(Details));
+    }
+
+    private async Task<ContentItem> GetUserContentItemAsync(User user, string contentType)
```
Psichorex, line 130: You can user `IEntity user` in parameters instead of the exact type `User`

src/Modules/OrchardCore.Commerce/Extensions/HttpContextExtensions.cs
```diff
@@ -12,6 +12,9 @@ public static class HttpContextExtensions
     public static Task<UserAddressesPart> GetUserAddressAsync(this HttpContext context) =>
         context.RequestServices.GetRequiredService<UserManager<IUser>>().GetUserAddressAsync(context.User);
 
+    public static Task<UserDetailsPart> GetUserDetailsAsync(this HttpContext context) =>
+        context.RequestServices.GetRequiredService<UserManager<IUser>>().GetUserDetailsAsync(context.User);
```
Psichorex, line 16: Would be nicer to break down this line before every `.`
Also there are many elements to this chain of calls and I can't see any null checks or possible breaks in the call.
For example it is possible for `.GetRequiredService` to throw `InvalidOperationException`

Psichorex (changes requested):

src/Modules/OrchardCore.Commerce/Controllers/UserController.cs
porgabi: What'd be the advantage of that though? This is only ever used in this controller where it's only used with the `User` type.

Psichorex: Just the convention to use interfaces for parameters rather than exact classes but if it's only used and will be used at one place we can go with `User˙user`

src/Modules/OrchardCore.Commerce/Extensions/HttpContextExtensions.cs
porgabi: IMO it's short enough to stay on one line.

I think `UserManager` is pretty much guaranteed to be found in all use cases of these extension methods, so it strikes me as unnecessary to add further null-checking here. This also seems to be in line with how `GetRequiredService()` is used in other solutions like OSOCE.

Psichorex: I just don't like stacking calls after eachother no matter how long they are to be honest.
We can have chain without any further checks if there are more examples of this.

src/Modules/OrchardCore.Commerce/Controllers/UserController.cs
DAud-IcI: afaik that convention applies to public/protected, but not private. For example you can see the similar analyzer rule regarding collections: https://github.com/meziantou/Meziantou.Analyzer/blob/main/docs/Rules/MA0016.md

Psichorex: Well gathering information is always beneficial. Now I know this aswell.

src/Modules/OrchardCore.Commerce/Extensions/HttpContextExtensions.cs
porgabi: For example, it's used similarly in [Helpful-Libraries/EnvironmentHttpContextExtensions](https://github.com/Lombiq/Helpful-Libraries/blob/1687c7b510ada082ea39c8f91451639b658ed552/Lombiq.HelpfulLibraries.AspNetCore/Extensions/EnvironmentHttpContextExtensions.cs#L17).

If you mean the following styling, IMO it uses too much vertical space and not enough horizontal. As far as I have seen, we most commonly put these on one line as long as they don't get too long.


```
context
    .RequestServices
    .GetRequiredService<UserManager<IUser>>()
    .GetUserDetailsAsync(context.User);
```

src/Modules/OrchardCore.Commerce/Controllers/UserController.cs
```diff
@@ -86,12 +86,53 @@ public async Task<IActionResult> AddressesPost()
...
+        if (_updateModelAccessor.ModelUpdater.ModelState.IsValid)
+        {
+            user.Put(UserDetails, userDetails);
+            _session.Save(user);
+            await _notifier.SuccessAsync(H["Your details have been updated."]);
+        }
+        else
+        {
+            var errors = _updateModelAccessor.ModelUpdater.GetModelErrorMessages();
+            foreach (var error in errors.WhereNot(string.IsNullOrEmpty)) await _notifier.ErrorAsync(H[error]);
```
DAud-IcI, line 124: I'm pretty sure the validation errors are already localized so this should rather be `H["{0}", error]`.

src/Modules/OrchardCore.Commerce/Extensions/UserManagerExtensions.cs
```diff
@@ -12,8 +12,16 @@ namespace Microsoft.AspNetCore.Identity;
...
     /// </summary>
     public static async Task<UserAddressesPart> GetUserAddressAsync(this UserManager<IUser> userManager, ClaimsPrincipal principal) =>
         await userManager.GetUserAsync(principal) is User user ? user.As<ContentItem>(UserAddresses)?.As<UserAddressesPart>() : null;
+
+    /// <summary>
+    /// Returns the custom user setting of <see cref="UserDetailsPart"/> for the given <paramref name="principal"/>
+    /// if the corresponding <see cref="User"/> is not <see langword="null"/>, returns <see langword="null"/> otherwise.
+    /// </summary>
+    public static async Task<UserDetailsPart> GetUserDetailsAsync(this UserManager<IUser> userManager, ClaimsPrincipal principal) =>
+        await userManager.GetUserAsync(principal) is User user ? user.As<ContentItem>(UserDetails)?.As<UserDetailsPart>() : null;
```
DAud-IcI, line 26: If you write `(await userManager.GetUserAsync(principal) as User)?` instead of `user` then the conditional expression can be avoided.

test/OrchardCore.Commerce.Tests.UI/Tests/UserTests/UserPersistenceTests.cs
```diff
@@ -116,29 +119,55 @@ public UserPersistenceTests(ITestOutputHelper testOutputHelper)
...
+                await context.ClickAndFillInWithRetriesAsync(
+                    By.Id("UserDetailsPart_VatNumber_Text"),
+                    VatNumber);
+                await context.ClickReliablyOnAsync(By.XPath("//label[contains(., 'User is a corporation')]"));
+
+                await context.ClickReliablyOnSubmitAsync();
+                context.ShouldBeSuccess("Your details have been updated.");
 
+                const string getInputsScript = @"return JSON.stringify(
+                    Array.from(document.querySelectorAll('form[action=\'/user/details\'] input, form[action=\'/user/details\'] select'))
```
DAud-IcI, line 141: You can also use backticks as string delimiter in JS (even if it's not a template string), so you won't run out of quote types:
```suggestion
                    Array.from(document.querySelectorAll(`form[action='/user/details'] input, form[action='/user/details'] select`))
```

src/Modules/OrchardCore.Commerce/Extensions/HttpContextExtensions.cs
Psichorex: Yes I was thinking about the above styling.
Yes, as you say it doesn't interfere with our 120-150 char length rule but breaking down the lines in my opinion makes the chain so much more readable. 
But this is "just" styling and one line so I consider this case resolved by leaving it as 1liner.

src/Modules/OrchardCore.Commerce/Extensions/UserManagerExtensions.cs
Psichorex: `(await userManager.GetUserAsync(principal) as User).As<ContentItem>(UserDetails)?.As<UserDetailsPart>();` can here indeed be a shortcut as you don't need the instance of the `User` type in this class.

Psichorex: Same applies for the method above this one.

Psichorex (approved):

PR #315, opened 2023-07-15
OCC-163: PriceVariantsPart doesn't support InventoryPart

### LINKED ISSUE #302: PriceVariantsPart doesn't support InventoryPart (OCC-163)
### Describe the bug

It should track the amount for each variant separately.

### To Reproduce

1. Create a content type with PriceVariantsPart and InventoryPart
2. Buy any variant.
3. The single stock number reduces.

### Expected behavior

Should be multiple numbers for each variant. Somehow.

[Jira issue](https://lombiq.atlassian.net/browse/OCC-163)

porgabi:
OCC-163

Fixes #302


src/Modules/OrchardCore.Commerce.Inventory/Drivers/InventoryPartDisplayDriver.cs
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    // Despite the Clear() calls inside UpdateAsync(), the Inventory property retains its old values along with the
+    // new ones, hence the filtering below.
+    private static void BuildViewModel(InventoryPartViewModel model, InventoryPart part)
+    {
+        var inventory = part.Inventory ?? new Dictionary<string, int>();
+        if (inventory.Any())
+        {
+            // Workaround for InventoryPart storing the outdated inventory entries along with the updated ones.
```
porgabi, line 79: The `InventoryKeys` list property is used as a workaround to the incorrectly working dictionary property. For some reason, list properties do not retain cleared values (which is working as intended), while dictionaries do.

src/Modules/OrchardCore.Commerce.Inventory/Views/InventoryPart.Edit.cshtml
```diff
@@ -0,0 +1,19 @@
+@model InventoryPartViewModel
+
+<h3>@T["Inventories"]</h3>
+<div class="row">
+    @foreach (var inventoryKey in Model.Inventory.Keys)
+    {
+        var inventoryTitle = Model.Inventory.Count > 1 ? inventoryKey.Split('-').Last() : string.Empty;
```
Psichorex, line 7: It would be reasonable to add a `SplitByDash` string extension instead of using `.Split('-')` like 8 times in the solution.

src/Modules/OrchardCore.Commerce/Drivers/ProductPartDisplayDriver.cs
```diff
@@ -48,22 +52,74 @@ public class ProductPartDisplayDriver : ContentPartDisplayDriver<ProductPart>
 
         part.Sku = part.Sku.ToUpperInvariant();
 
+        if (part.ContentItem.As<InventoryPart>() is { } inventoryPart)
+        {
+            part.CanBeBought.Clear();
+
+            var correctInventory = inventoryPart.Inventory
```
Psichorex, line 59: This invertory correction could be extraced into a method and use that in multiple places.

src/Modules/OrchardCore.Commerce/Views/TextProductAttributeField.cshtml
```diff
@@ -6,6 +6,12 @@
     var settings = attr?.Settings as TextProductAttributeFieldSettings;
     var id = $"{Model.IdPrefix}-attribute-{index}";
     var optionIndex = 0;
+
+    var sku = (string)Model.Sku;
```
Psichorex, line 10: I would use `Model.Sku as string` as this will return null if the cast fails.
Ir it is intentional to have an exception if it fails then I am okay with this.

src/Modules/OrchardCore.Commerce/Views/TextProductAttributeField.cshtml
```diff
@@ -6,6 +6,12 @@
...
+    var canBeBoughtDictionary = ((Dictionary<string, bool>)Model.CanBeBought);
```
Psichorex, line 11: Same can happen here (T)item cast can lead to an exception.

src/Modules/OrchardCore.Commerce/Views/TextProductAttributeField.cshtml
```diff
@@ -6,6 +6,12 @@
...
+    var availableVariants = canBeBoughtDictionary.Where(keyValuePair => keyValuePair.Value == true);
```
Psichorex, line 12: `canBeBoughtDictionary` might return null thus a null check would be welcome here for safety reasons.

Psichorex (changes requested):

src/Modules/OrchardCore.Commerce/Views/TextProductAttributeField.cshtml
porgabi: Realistically, SKU should never be empty on ProductPart, but I'll use the `as string` approach for consistency.

src/Modules/OrchardCore.Commerce.Inventory/Views/InventoryPart.Edit.cshtml
porgabi: The question is where to add this? `Split('-')` is used in 3 different projects, but none of these are suitable to place this extension method as you couldn't make use of it in all places. For example, if you create it in `OrchardCore.Commerce`, you cannot use it in `OrchardCore.Inventory` and `OrchardCore.MoneyDataType`.

To be honest, it's a simple and short enough call to just have it as is, but if you have a suggestion that'd allow such an extension method to be used in all solutions in the project in a simple manner, I'll add it.

Psichorex: I find your argument sufficient to discard my request.

Psichorex (approved):

PR #321, opened 2023-07-26
OCC-52: No Payment Needed Checkout

### LINKED ISSUE #13: No Payment Needed Checkout (OCC-52)
Implementation should allow a user to checkout without a payment if their shopping cart totals no cost. 

[Jira issue](https://lombiq.atlassian.net/browse/OCC-52)

porgabi:
[OCC-52](https://lombiq.atlassian.net/browse/OCC-52)

Fixes #13


src/Modules/OrchardCore.Commerce/Services/StripePaymentService.cs
```diff
@@ -170,6 +176,11 @@ public async Task UpdateOrderToOrderedAsync(PaymentIntent paymentIntent)
             await workflowManager.TriggerEventAsync(nameof(OrderCreatedEvent), order, "Order-" + order.ContentItemId);
         }
 
+        var currentShoppingCart = await _shoppingCartPersistence.RetrieveAsync();
+
+        // Decrease inventories of purchased items.
+        await _productInventoryService.UpdateInventoriesAsync(currentShoppingCart.Items);
```
porgabi, line 182: Moved this call here to address bug where inventories were decreased even after unsuccessful payment.

src/Modules/OrchardCore.Commerce/Views/Checkout.cshtml
```diff
@@ -11,7 +13,7 @@
 
 <section id="checkout">
     <div class="payment-form_container">
-        <form id="payment-form" class="payment-form">
+        <form asp-action="@nameof(PaymentController.CheckoutWithoutPayment)" asp-route-shoppingCartId="@Model.ShoppingCartId" id="payment-form" class="payment-form">
```
porgabi, line 16: When the Stripe payment form is rendered, this form is not submitted, so it doesn't interfere with payment checkouts.

src/Modules/OrchardCore.Commerce/Abstractions/IStripePaymentService.cs
```diff
@@ -31,9 +31,15 @@ public interface IStripePaymentService
...
+    /// Creates an order content item without payment in the database based on the current <see cref="ShoppingCart"/> content.
     /// </summary>
-    Task UpdateOrderToOrderedAsync(PaymentIntent paymentIntent);
+    Task<ContentItem> CreateNoPaymentOrderFromShoppingCartAsync(IUpdateModelAccessor updateModelAccessor);
+
+    /// <summary>
+    /// Updates the corresponding order status to Ordered for the given <paramref name="paymentIntent"/> or the
+    /// provided <paramref name="orderItem"/>.
+    /// </summary>
+    Task UpdateOrderToOrderedAsync(PaymentIntent paymentIntent = null, ContentItem orderItem = null);
```
DAud-IcI, line 42: Please use `<see cref="OrderStatuses.Ordered"/>`

src/Modules/OrchardCore.Commerce/Abstractions/IStripePaymentService.cs
```diff
@@ -31,9 +31,15 @@ public interface IStripePaymentService
     Task<ContentItem> CreateOrUpdateOrderFromShoppingCartAsync(PaymentIntent paymentIntent, IUpdateModelAccessor updateModelAccessor);
 
     /// <summary>
-    /// Updates the corresponding order status to Order for the given <paramref name="paymentIntent"/>.
+    /// Creates an order content item without payment in the database based on the current <see cref="ShoppingCart"/> content.
     /// </summary>
-    Task UpdateOrderToOrderedAsync(PaymentIntent paymentIntent);
+    Task<ContentItem> CreateNoPaymentOrderFromShoppingCartAsync(IUpdateModelAccessor updateModelAccessor);
```
DAud-IcI, line 36: It's not clear why is this in the Stripe payment service? It's not Stripe-specific. I think it should be in the `IPaymentService` so it won't break when we incorporate the option for other payment providers.
(This method is different from `UpdateOrderToOrderedAsync` that actually Stripe stuff, which is fine for now.)

DAud-IcI, line 36: Why do you pass in the `IUpdateModelAccessor` as an argument? It should be provided to the service via dependency injection. (that's what already happens in the `PaymentController` where you call it)

src/Modules/OrchardCore.Commerce/Controllers/PaymentController.cs
```diff
@@ -166,6 +166,19 @@ public async Task<IActionResult> Success(string orderId)
         return View(order);
     }
 
+    [AllowAnonymous]
+    [HttpPost]
+    [ValidateAntiForgeryToken]
+    public async Task<IActionResult> CheckoutWithoutPayment()
+    {
+        var order = await _stripePaymentService.CreateNoPaymentOrderFromShoppingCartAsync(_updateModelAccessor);
```
DAud-IcI, line 174: The `CreateNoPaymentOrderFromShoppingCartAsync` can return null too, and then this will cause an NRE.
You should do a null check and return `NotFound`.

src/Modules/OrchardCore.Commerce/Services/StripePaymentService.cs
```diff
@@ -278,6 +286,49 @@ public async Task UpdateOrderToPaymentFailedAsync(PaymentIntent paymentIntent)
...
+
+        var guid = Guid.NewGuid().ToString();
+        order.DisplayText = T["Order {0}", guid];
+
+        var lineItems = await CreateOrderLineItemsAsync(currentShoppingCart);
+
+        var cartViewModel = await _shoppingCartHelpers.CreateShoppingCartViewModelAsync(
+            shoppingCartId: null,
+            order.As<OrderPart>().ShippingAddress.Address,
+            order.As<OrderPart>().BillingAddress.Address);
```
DAud-IcI, line 307: After the model is created here, check if the `cartViewModel.Totals` is empty or all elements add up to zero. Return null if that's not the case. I doubt that would ever happen, but if it somehow did (e.g. via a bugged third party promotion provider or some other wacky stuff) that could be very costly.

src/Modules/OrchardCore.Commerce/Services/StripePaymentService.cs
```diff
@@ -278,6 +286,49 @@ public async Task UpdateOrderToPaymentFailedAsync(PaymentIntent paymentIntent)
...
+
+        order.Alter<OrderPart>(orderPart =>
+        {
+            // Shopping cart
+            orderPart.LineItems.Clear();
+            orderPart.LineItems.AddRange(lineItems);
```
DAud-IcI, line 313: We have an extension method for this in HL called `ICollection<T>.SetItems(newValues)`. (please change in in `CreateOrUpdateOrderFromShoppingCartAsync` and `OrderPartDisplayDriver.UpdateAsync` as well)

src/Modules/OrchardCore.Commerce/Services/StripePaymentService.cs
```diff
@@ -278,6 +286,49 @@ public async Task UpdateOrderToPaymentFailedAsync(PaymentIntent paymentIntent)
...
+        var currentShoppingCart = await _shoppingCartPersistence.RetrieveAsync();
+
+        var order = await _contentManager.NewAsync(Constants.ContentTypes.Order);
+        if (await UpdateOrderWithDriversAsync(order, updateModelAccessor))
+        {
+            return null;
+        }
+
+        var guid = Guid.NewGuid().ToString();
+        order.DisplayText = T["Order {0}", guid];
```
DAud-IcI, line 300: I see this recurring in two other places too. I think it should be put into an OrderPart content Handler. (in the format you can see in `PaymentService.FinalModificationOfOrderAsync()`)

src/Modules/OrchardCore.Commerce/Services/StripePaymentService.cs
```diff
@@ -278,6 +286,49 @@ public async Task UpdateOrderToPaymentFailedAsync(PaymentIntent paymentIntent)
...
+        order.Alter<OrderPart>(orderPart =>
+        {
+            // Shopping cart
+            orderPart.LineItems.Clear();
+            orderPart.LineItems.AddRange(lineItems);
+
+            orderPart.OrderId.Text = guid;
+            orderPart.Status = new TextField { ContentItem = order, Text = OrderStatuses.Pending.HtmlClassify() };
+
+            // Store the current applicable discount info so they will be available in the future.
```
DAud-IcI, line 318:
```suggestion
            // Store the current applicable discount info, so they will be available in the future.
```

src/Modules/OrchardCore.Commerce/Services/StripePaymentService.cs
```diff
@@ -278,6 +286,49 @@ public async Task UpdateOrderToPaymentFailedAsync(PaymentIntent paymentIntent)
...
+            order.As<OrderPart>().BillingAddress.Address);
+
+        order.Alter<OrderPart>(orderPart =>
+        {
+            // Shopping cart
+            orderPart.LineItems.Clear();
+            orderPart.LineItems.AddRange(lineItems);
+
+            orderPart.OrderId.Text = guid;
+            orderPart.Status = new TextField { ContentItem = order, Text = OrderStatuses.Pending.HtmlClassify() };
```
DAud-IcI, line 316: You don't have to overwrite the whole field. `OrderPart.Status` is never null because it has a `new()` default value, so it's enough to just overwrite `orderPart.Status.Text`.

src/Modules/OrchardCore.Commerce/Views/Checkout.cshtml
```diff
@@ -89,12 +91,26 @@
                 }
             }
 
-            <shape type="StripeCheckout"
-                   prop-DefaultTotal="@Model.SingleCurrencyTotal"
-                   prop-NetTotal="@Model.NetTotal"
-                   prop-GrossTotal="@Model.GrossTotal"
-                   prop-PublishableKey="@Model.StripePublishableKey"
-                   prop-ClientSecret="@Model.PaymentIntentClientSecret"/>
+            @*If totals are zero, there's no need for payment.*@
```
DAud-IcI, line 94:
```suggestion
            @* If totals are zero, there's no need for payment. *@
```

DAud-IcI (changes requested):

src/Modules/OrchardCore.Commerce/Abstractions/IStripePaymentService.cs
porgabi: While at it, I also moved `CreateOrderLineItemsAsync()` to `IPaymentService` as it seems better to have `StripePaymentService` depend on `PaymentService` than vice-versa.

src/Modules/OrchardCore.Commerce/Handlers/OrderPartHandler.cs
```diff
@@ -0,0 +1,34 @@
...
+    public OrderPartHandler(IStringLocalizer<OrderPartHandler> stringLocalizer, ISession session)
+    {
+        _session = session;
+        T = stringLocalizer;
+    }
+
+    public override Task UpdatedAsync(UpdateContentContext context, OrderPart part)
+    {
+        var guid = string.IsNullOrEmpty(part.OrderId.Text) ? Guid.NewGuid().ToString() : part.OrderId.Text;
+        part.OrderId.Text = guid;
```
hishamco:
```suggestion
        part.OrderId.Text = part.OrderId.Text ?? Guid.NewGuid().ToString();
```

src/Modules/OrchardCore.Commerce/Services/StripePaymentService.cs
porgabi: I added a handler that contains an `UpdatedAsync()` override; I chose to override this method as it gets called after both creating and updating an existing content item. Let me know if this is not what you meant.

DAud-IcI (changes requested):

PR #323, opened 2023-08-02
OCC-173: Discount is not applied if there is also another not discounted product in the shopping cart

### LINKED ISSUE #322: Discount is not applied if there is also another not discounted product in the shopping cart (OCC-173)
### Describe the bug

> I have noticed there’s a bug with discounts that makes the discount effectively disappear when there’s another item in the cart that’s not discounted.
>
> For example, TestProduct is on a sale that makes it cost $0, while Pricy Product is not on sale at all. When both products are in the cart, TestProduct’s discount disappears, but if Pricy Product is removed from the cart, the discount is back.
>
> Untitled
>
> This doesn’t seem to be the case when both products are on sale, so I assume it’s something to do with the combination of an on-sale product and a full-price product in a cart:
>
> image

### To Reproduce

1. Create a new product with discount.
2. Add it to cart.
3. Add the TestProduct to cart as well.
4. Go to the cart.
5. See that the discounted product is seemingly not discounted and the total is according to the full price.

### Expected behavior

Discount should apply on individual products.

[Jira issue](https://lombiq.atlassian.net/browse/OCC-173)

porgabi:
[OCC-173](https://lombiq.atlassian.net/browse/OCC-173)

Fixes #322

### DIFF
--- a/test/OrchardCore.Commerce.Tests.UI/Tests/PromotionTests/PromotionBehaviourTests.cs
+++ b/test/OrchardCore.Commerce.Tests.UI/Tests/PromotionTests/PromotionBehaviourTests.cs
@@ -56,10 +56,16 @@ public class PromotionBehaviourTests : UITestBase
                 // Discount should appear in cart when discounted product is the only item in the cart.
                 context.Get(By.CssSelector(".shopping-cart-table-unit-price")).Text.ShouldBe(DiscountedPrice);
 
+                // Old price should be shown in cart.
+                context.Get(By.CssSelector(".shopping-cart-table-old-price > del")).Text.ShouldBe(FullPrice);
+
                 // Discount should appear in cart when discounted product is not the only item in the cart.
                 await context.GoToRelativeUrlAsync("/testproduct");
                 await context.ClickReliablyOnSubmitAsync();
                 context.Get(By.CssSelector(".shopping-cart-table-unit-price")).Text.ShouldBe(DiscountedPrice);
+
+                // Old price should be shown for discounted product in cart when the cart contains a full-price item too.
+                context.Get(By.CssSelector(".shopping-cart-table-old-price > del")).Text.ShouldBe(FullPrice);
 
                 // Total should reflect discount as well.
                 context.Get(By.CssSelector(".shopping-cart-table-totals > div")).Text.ShouldBe("$8.00");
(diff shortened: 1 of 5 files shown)


test/OrchardCore.Commerce.Tests.UI/Tests/PromotionTests/PromotionBehaviourTests.cs
```diff
@@ -35,4 +36,33 @@ async Task CheckDiscountPresenceAsync(bool hasNewPrice)
...
+
+    [Theory, Chrome]
+    public Task DiscountShouldProperlyAppearOnDiscountedProduct(Browser browser) =>
+        ExecuteTestAfterSetupAsync(
+            async context =>
+            {
+                // Discount should appear on discounted product's page.
+                await context.GoToRelativeUrlAsync("/testdiscountedproduct");
+                context.Get(By.CssSelector(".price-part-price-field-value > del")).Text.ShouldBe("$5.00");
+                context.Get(By.CssSelector(".price-part-price-field-value > span")).Text.ShouldBe("$3.00");
```
DemeSzabolcs, line 48: I think these could be move to separate variables, since they are used multiple times. 
`.price-part-price-field-value`
`$3.00`
`$5.00`

test/OrchardCore.Commerce.Tests.UI/Tests/PromotionTests/PromotionBehaviourTests.cs
```diff
@@ -35,4 +36,33 @@ async Task CheckDiscountPresenceAsync(bool hasNewPrice)
...
+                await context.ClickReliablyOnSubmitAsync();
+
```
DemeSzabolcs, line 50: Please also check for the Old Price on the cart page, since it seems like it's still not displayed: 
oldprice

DemeSzabolcs (changes requested):

test/OrchardCore.Commerce.Tests.UI/Tests/PromotionTests/PromotionBehaviourTests.cs
porgabi: That's a separate problem, but I'll take a look if there's a quick fix.

DemeSzabolcs: Ah okay. I didn't see an issue about this though. So if you want you can fix it here, or create an issue.

porgabi: Addressed it here.

DemeSzabolcs (approved):

PR #327, opened 2023-08-18
OCC-65: Make the line items editable in the OrderPart editor

### LINKED ISSUE #117: Make the line items editable in the OrderPart editor
https://github.com/OrchardCMS/OrchardCore.Commerce/blob/e9dba990fbf2359dffa5af31126a800e59ad4118/OrchardCore.Commerce/Views/OrderPart_Edit.cshtml#L4
 (This issue was created as part of the effort to externalize TODOs from the code base.)

porgabi:
[OCC-65](https://lombiq.atlassian.net/browse/OCC-65)

Fixes #117

### DIFF
--- a/src/Modules/OrchardCore.Commerce/Extensions/ProductServiceExtensions.cs
+++ b/src/Modules/OrchardCore.Commerce/Extensions/ProductServiceExtensions.cs
@@ -18,5 +18,5 @@ public static class ProductServiceExtensions
     public static async Task<IDictionary<string, ProductPart>> GetProductDictionaryByContentItemVersionsAsync(
         this IProductService service,
         IEnumerable<string> contentItemVersions) =>
-        (await service.GetProductsByContentItemVersionsAsync(contentItemVersions)).ToDictionary(product => product.Sku);
+        (await service.GetProductsByContentItemVersionsAsync(contentItemVersions.Distinct())).ToDictionary(product => product.Sku);
 }
--- a/src/Modules/OrchardCore.Commerce/Views/OrderPart_Edit.cshtml
+++ b/src/Modules/OrchardCore.Commerce/Views/OrderPart_Edit.cshtml
@@ -14,7 +14,7 @@
 @inject IPredefinedValuesProductAttributeService AttributesService
 
 @{
-    var availableCurrencyCodes = MoneyService.Currencies.Select(currency => currency.CurrencyIsoCode);
+    var serializedAvailableCurrencyCodes = JsonSerializer.Serialize(MoneyService.Currencies.ToDictionary(currency => currency.CurrencyIsoCode, currency => currency.Symbol));
     var defaultCurrencyCode = MoneyService.DefaultCurrency.CurrencyIsoCode ?? "USD";
 
     // Get all predefined attributes in the system by filtering for Products that have PriceVariantsPart attached.
@@ -41,7 +41,7 @@
         }
     }
 
-    var serializedDictionary = JsonSerializer.Serialize(availableAttributes);
+    var serializedAttributesDictionary = JsonSerializer.Serialize(availableAttributes);
 }
 
 <section id="shopping-cart">
@@ -89,11 +89,8 @@
                     </div>
                     <div class="col-xs-12 col-sm-1 d-flex">
                         <select :name="'OrderPart.LineItems[' + index + '].UnitPriceCurrencyIsoCode'" v-model="lineItem.unitPriceCurrencyIsoCode">
-                            @foreach (var currencyCode in availableCurrencyCodes)
-                            {
-                                <option value="@currencyCode">@currencyCode</option>
-                            }
-                        </select>
+                            <option v-for="(value, key) in availableCurrencyCodes" :value="key">{{ key }}</option>
+                       </select>
                     </div>
                     <div class="col-xs-12 col-sm-2 d-flex">
                         <div v-for="(values, key) in lineItem.availableAttributes">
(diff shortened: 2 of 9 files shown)


src/Modules/OrchardCore.Commerce/Views/OrderPart_Edit.cshtml
```diff
@@ -1,63 +1,122 @@
...
+    foreach (var priceVariant in allPriceVariantProducts)
     {
-        <div class="container">
-            <div class="d-none d-sm-flex row border-bottom mt-3 pb-2 font-weight-bold">
-                <div class="col-xs-12 col-sm-2">@T["Quantity"]</div>
-                <div class="col-xs-12 col-sm-4">@T["Product"]</div>
-                <div class="col-xs-12 col-sm-4">@T["Attributes"]</div>
-                <div class="col-xs-12 col-sm-1">@T["Price"]</div>
-            </div>
+        var attributes = AttributesService.GetProductAttributesRestrictedToPredefinedValues(priceVariant);
```
porgabi, line 27: One issue with this is that it only retrieves attributes when there already is a Price Variant Product among the Order's line items. Makes no sense to me as the query above specifically gets all Price Variant Products in the system regardless of what the Order contains.

This means that when the Order does not contain any Price Variant Products, attribute dropdowns are not available until a Price Variant Product is added. While this is sub-optimal, it also doesn't feel like a deal-breaker to me in terms of functionality.

src/Modules/OrchardCore.Commerce/Extensions/ProductServiceExtensions.cs
```diff
@@ -18,5 +18,5 @@ public static class ProductServiceExtensions
     public static async Task<IDictionary<string, ProductPart>> GetProductDictionaryByContentItemVersionsAsync(
         this IProductService service,
         IEnumerable<string> contentItemVersions) =>
-        (await service.GetProductsByContentItemVersionsAsync(contentItemVersions)).ToDictionary(product => product.Sku);
+        (await service.GetProductsByContentItemVersionsAsync(contentItemVersions.Distinct())).ToDictionary(product => product.Sku);
```
porgabi, line 21: This is a fix for the following:
> I noticed Price Variant Products aren't listed in the Order editor after making a purchase (when there are multiple different variants purchased at the same time, at least). An Order is created, but with no charges or products listed.

src/Modules/OrchardCore.Commerce/Views/OrderPart_Edit.cshtml
DAud-IcI: It's not clear to me why you need `allPriceVariantProducts`. I've added a text product attribute field with predefined values to Product, removed the `.Where(product => product.As<PriceVariantsPart>() != null)` from above, similarly use `productPart.ContentItem` in `HandleSelectedAttributes` and it seems to work:
image

DAud-IcI (changes requested):

src/Modules/OrchardCore.Commerce/Views/OrderPart_Edit.cshtml
porgabi: I was under the impression only Price Variant Products need attributes as their prices are based on those. Is that not the purpose of attributes? Should any Product be able to have attributes?

src/Modules/OrchardCore.Commerce/Views/OrderPart_Edit.cshtml
```diff
@@ -1,63 +1,122 @@
...
         </div>
-    }
-    else
-    {
-        <p>@T["This order is empty."]</p>
-    }
+        <div class="row mt-3 mb-3 pb-2">
+            <div class="col-xs-12 col-sm-10 font-weight-bold">@T["Total"]</div>
+            <div class="col-xs-12 col-sm-1">
+                <div>@Model.Total</div>
```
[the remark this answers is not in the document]
porgabi: In addition, I also made the displayed currency symbol dynamic, currently controlled by the topmost Product's selected currency. Thought it'd make sense to have it dynamic now that the total value is.

Let me know if you think there's a better way to determine which symbol to display when there are multiple Products. I figured this'd do as all selected currencies need to be the same anyway.

image

src/Modules/OrchardCore.Commerce/Views/OrderPart_Edit.cshtml
```diff
@@ -185,16 +185,29 @@
...
+            lineItems: lineItems,
+        },
+        computed: {
+            calculatedTotal() {
+                return this.lineItems
+                    .map(item => item.quantity * item.unitPriceValue)
+                    .reduce((total, value) => total + value, 0);
+            },
+            displayedCurrency() {
+                return this.lineItems.map(item => this.availableCurrencyCodes[item.unitPriceCurrencyIsoCode])[0];
```
DAud-IcI, line 208: Why map the whole list? I think you want [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) (`?.` operator).

porgabi: > Why map the whole list?

To cook is to sometimes get lost in the sauce. Thanks, should be simpler now.

src/Modules/OrchardCore.Commerce/Views/OrderPart_Edit.cshtml
```diff
@@ -1,63 +1,122 @@
...
+        <h3>@T["Ordered Items"]</h3>
+    </div>
+    <fieldset class="mb-3">
+        <label>@T["Product properties"]</label>
+        <ul>
+            <li class="hint">
+                @T["Currencies: Selected currencies need to be the same for all Products."]
+            </li>
+            <li class="hint">
+                @T["Attributes: Attributes are only to be selected for Price Variant Products. Only select a value for those attributes that are relevant to the product in question."]
```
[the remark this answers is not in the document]
porgabi: > Why are you able to select irrelevant attributes?

This was intended as a way to allow adding attributes to a Product that is being added as part of the current request, but it's irrelevant now as attributes are only populated for a Product after it's been added. Same for the mixed-together attribute values.

> It would be better if availableAttributes was Dictionary<string, Dictionary<string, List<string>>> so you get available attributes separately by SKU. Then you only list the applicable ones by lineItem.productSku.

Adjusted it to be this way.

PR #330, opened 2023-08-23
OCC-137: Add Shopping Cart Loaded Workflow Event

### LINKED ISSUE #269: Add Shopping Cart Loaded Workflow Event (OCC-137)

Currently there is no way to add automatic entries to the shopping cart, such as **shipping cost**. Since this kind of stuff is tied to the business logic, we should support it via workflows so the custom logic can be coded in admin using Jint/Javascript.

- Add a new method for when the shopping cart is loaded (e.g. `Task LoadedAsync(string shoppingCartId, ShoppingCart shoppingCart)`). This should be invoked in `IShoppingCartPersistence.RetrieveAsync()`, so it behaves the same way every time the shopping cart is loaded.
- Create workflow triggers for each method of `IShoppingCartEvents`
- Create a `WorkflowShoppingCartEvents : IShoppingCartEvents` service that triggers each workflow. 
  - It should not inherit from `ShoppingCartEventsBase` so we won't forget to add workflow triggers whenever a new method is added to `IShoppingCartEvents`.
- If [`TableField` is implemented in Lombiq.Datatables](https://github.com/Lombiq/Orchard-Data-Tables/issues/133), then we can suggest using it via the _Retrieve Content_ task to store additional tabular data that may be needed to calculate shipping costs.

We could create a specific implementation for specific features like adding shipping cost, but it would be too inflexible.

[Jira issue](https://lombiq.atlassian.net/browse/OCC-137)

DAud-IcI:
[OCC-137](https://lombiq.atlassian.net/browse/OCC-137)
Fixes #269

### DIFF
--- a/docs/features/workflows.md
+++ b/docs/features/workflows.md
@@ -11,18 +11,24 @@ All of these workflows expect to return one or more outputs which is passed back
 
 ### "Cart displaying" Event
 
+Executes after the shopping cart data is prepared, but before the shapes are rendered.
+
 - Input: `ShoppingCartDisplayingEventContext` object containing the current shopping cart's headers and lines.
-- Outputs: either inputs are optional.
+- Outputs: either outputs are optional.
   - Headers: `LocalizedHtmlString` array. The shopping cart header labels in order. If you have to support multiple locales, make sure to use the object format mentioned above, because `LocalizedHtmlString.Name` is used to generate the template name for the corresponding shopping cart column's cells.
   - Lines: `ShoppingCartLineViewModel` array. This is only for display, in most cases you shouldn't have to return this output.
 
 ### "Verifying cart item" Event
+
+Executes before an item is added to the shopping cart to check whether it can be added based on inventory status.
 
 - Input: `ShoppingCartItem` object.
 - Outputs:
   - Error: `LocalizedHtmlString` or `null`. The error message to be displayed if the input item can't be added to the cart. You can simply not output anything if the validation passes.
 
 ### "Cart loaded" Event
+
+Executes after the shopping cart content is loaded from the store and before it's displayed or used for calculation.
 
 - Input: `ShoppingCart` object.
 - Outputs:
--- a/docs/features/workflows.md
+++ b/docs/features/workflows.md
@@ -5,7 +5,7 @@
 These events get triggered by `WorkflowShoppingCartEvents` which implements the `IShoppingCartEvents` interface. For each you can access the input as a .NET object using the `Context` workflow input and the serialized version as the `JSON` workflow input.
 All of these workflows expect to return one or more outputs which is passed back to the invoking code.
 
-> ⚠ If you want to return an altered version of the input as the output, please always just the JSON which is already serialized in the expected format used by OrchardCore.Commerce's converters. For example you can use the JS experssion `JSON.parse(input('JSON'))`.
+> ⚠ If you want to return an altered version of the input as the output, please always just the JSON which is already serialized in the expected format used by OrchardCore.Commerce's converters. For example you can use the JS expression `JSON.parse(input('JSON'))`.
 
 > ℹ When your output contains `LocalizedHtmlString`, it can be represented in JS either as `string` or `{ Name: string, Value: string }`. In case of just `string` the same text becomes `LocalizedHtmlString.Name` and `LocalizedHtmlString.Value` too.
 
--- a/docs/features/workflows.md
+++ b/docs/features/workflows.md
@@ -1,0 +1,47 @@
+# Workflows
+
+## Shopping Cart Event Workflow Events
+
+These events get triggered by `WorkflowShoppingCartEvents` which implements the `IShoppingCartEvents` interface. For each you can access the input as a .NET object using the `Context` workflow input and the serialized version as the `JSON` workflow input.
+All of these workflows expect to return one or more outputs which is passed back to the invoking code.
+
+> ⚠ If you want to return an altered version of the input as the output, please always just the JSON which is already serialized in the expected format used by OrchardCore.Commerce's converters. For example you can use the JS experssion `JSON.parse(input('JSON'))`.
+
+> ℹ When your output contains `LocalizedHtmlString`, it can be represented in JS either as `string` or `{ Name: string, Value: string }`. In case of just `string` the same text becomes `LocalizedHtmlString.Name` and `LocalizedHtmlString.Value` too.
+
+### "Cart displaying" Event
+
+- Input: `ShoppingCartDisplayingEventContext` object containing the current shopping cart's headers and lines.
+- Outputs: either inputs are optional.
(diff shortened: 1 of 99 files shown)


src/Modules/OrchardCore.Commerce/Events/WorkflowShoppingCartEvents.cs
```diff
@@ -0,0 +1,104 @@
...
+            ["Context"] = input,
+            ["JSON"] = JsonConvert.SerializeObject(input),
+        };
+
+        // Start new workflows whose types have a corresponding starting activity.
+        var workflowTypesToStart = await _workflowTypeStore.GetByStartActivityAsync(name);
+        var contexts = new List<WorkflowExecutionContext>();
+        foreach (var workflowType in workflowTypesToStart)
+        {
+            var startActivity = workflowType.Activities.First(x => x.IsStart && x.Name == name);
```
DemeSzabolcs, line 78: Please use a more telling variable name (`x`).

DemeSzabolcs (changes requested):

src/Libraries/OrchardCore.Commerce.MoneyDataType/Serialization/LegacyAmountConverter.cs
```diff
@@ -78,4 +83,15 @@ public override void WriteJson(JsonWriter writer, Amount amount, JsonSerializer
 
         writer.WriteEndObject();
     }
+
+    public static Amount ReadString(string text)
+    {
+        var parts = text.Split();
+        if (parts.Length < 2) throw new InvalidOperationException($"Unable to parse string amount \"{text}\".");
```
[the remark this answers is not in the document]
DAud-IcI: Yeah, it was intentional. By doing `string.Join` with an empty separator we remove all spaces after the first anyway. So you can enter "HUF 1 000 000" and it will still parse. It's free ~real estate~ readability!

src/Modules/OrchardCore.Commerce/Controllers/ShoppingCartController.cs
```diff
@@ -109,13 +110,9 @@ public async Task<ActionResult> AddItem(ShoppingCartLineUpdateModel line, string
...
                 storeIfOk: true);
 
-            if (_workflowManager != null)
-            {
-                await _workflowManager.TriggerEventAsync(
-                    nameof(ProductAddedToCartEvent),
-                    new { LineItem = parsedLine },
-                    "ShoppingCart-" + _shoppingCartPersistence.GetUniqueCartId(shoppingCartId));
-            }
+            await _workflowManagers.TriggerEventAsync<ProductAddedToCartEvent>(
```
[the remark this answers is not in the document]
DAud-IcI: OrchardCore.Workflows is not required and we still use `IEnumerable<IWorkflowManager> _workflowManagers` instead of `IWorkflowManager _workflowManager`. The empty checking has been abstracted into the extension method [here](https://github.com/Lombiq/Helpful-Libraries/blob/4a0828dc015ec98f0501165b8f9d754b7522a783/Lombiq.HelpfulLibraries.OrchardCore/Workflow/WorkflowManagerExtensions.cs#L59) (which uses the new [`InvokeFirstOrCompletedAsync`](https://github.com/Lombiq/Helpful-Libraries/blob/4a0828dc015ec98f0501165b8f9d754b7522a783/Lombiq.HelpfulLibraries.Common/Extensions/EnumerableExtensions.cs#L392) extension method) to avoid code repetition.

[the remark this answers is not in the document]
DemeSzabolcs: I see, thanks!

src/Modules/OrchardCore.Commerce/Serialization/LegacyShoppingCartItemConverter.cs
```diff
@@ -0,0 +1,85 @@
...
+    {
+        writer.WriteStartObject();
+
+        writer.WritePropertyName(ShoppingCartItemConverter.QuantityName);
+        writer.WriteValue(value.Quantity);
+
+        writer.WritePropertyName(ShoppingCartItemConverter.SkuName);
+        writer.WriteValue(value.ProductSku);
+
+        if (value.Prices != null && value.Prices.Any())
```
DemeSzabolcs, line 24: Can't we do:
`value.Prices?.Any() == true`
here and below?

docs/features/workflows.md
```diff
@@ -0,0 +1,53 @@
...
+
+### "Verifying cart item" Event
+
+Executes before an item is added to the shopping cart to check whether it can be added based on inventory status.
+
+- Input: `ShoppingCartItem` object.
+- Outputs:
+  - Error: `LocalizedHtmlString` or `null`. The error message to be displayed if the input item can't be added to the cart. You can simply not output anything if the validation passes.
+
+### "Cart loaded" Event
```
DemeSzabolcs, line 29: I'm only navigating to 1 product page, but it seems like the Cart Loaded event triggers 3 times always: 
cart

This is always the case with the Cart Loaded event.

DemeSzabolcs (changes requested):

docs/features/workflows.md
DAud-IcI: It's triggered each time the `IShoppingCartPersistence.RetrieveAsync()` is invoked so that's not surprising. Once by the shopping cart widget, but what's odd is that it's triggered twice by different method calls in `ShoppingCartHelpers.EstimateProductAsync()` used by the `DiscountPartDisplayDriver` to determine if any discounts are eligible. The latter probably can be avoided by optimization (so only load once during estimation and don't estimate if there is no discount to apply as is the case in your example).

src/Modules/OrchardCore.Commerce/Serialization/LegacyShoppingCartItemConverter.cs
```diff
@@ -0,0 +1,85 @@
...
+        writer.WritePropertyName(ShoppingCartItemConverter.SkuName);
+        writer.WriteValue(value.ProductSku);
+
+        if (value.Prices?.Any() == true)
+        {
+            writer.WritePropertyName(ShoppingCartItemConverter.PricesName);
+            serializer.Serialize(writer, value.Prices);
+        }
+
+        if (value.Attributes != null && value.Attributes.Any())
```
DemeSzabolcs, line 30: This too could be:
`value.Prices?.Any() == true`

https://github.com/OrchardCMS/OrchardCore.Commerce/pull/330#discussion_r1306442721

DemeSzabolcs (changes requested):

src/Modules/OrchardCore.Commerce/Serialization/LegacyShoppingCartItemConverter.cs
DemeSzabolcs: @DAud-IcI Sorry it's `value.Attributes` not `Prices` here.

DemeSzabolcs (approved):

PR #356, opened 2023-09-26
OCC-188: Add a payment request screen for Orders with the "Pending" status

### LINKED ISSUE #345: Add a payment request screen for Orders with the "Pending" status (OCC-188)
Followup to #117

Currently we can create new Orders from the content item editor, but there is no way to have them paid from withing OCC. If you create a custom order it has to be paid outside of the site and then the order manually updated. 

Add a new action to the `PaymentController` (e.g. _~/checkout/request_) that displays the payment UI (e.g. the card widget from Stripe) so the payment can be fulfilled by the customer. Also add a button to the Order display view if the status is pending.

It could be possible to display the payment controls right in the Order display if needed, but I think that's too much complexity.
I also considered automatically redirecting the user to the payment request screen if the order is pending, but that may be confusing. Also it can be achieved using stock OC workflows if that's a desired behavior so there is no need to force it out of the box.

[Jira issue](https://lombiq.atlassian.net/browse/OCC-188)

porgabi:
[OCC-188](https://lombiq.atlassian.net/browse/OCC-188)

Fixes #345


src/Modules/OrchardCore.Commerce/Controllers/PaymentController.cs
```diff
@@ -150,6 +157,59 @@ public async Task<IActionResult> Validate()
...
+        {
+            return NotFound();
+        }
+
+        var currency = orderPart.LineItems[0].LinePrice.Currency;
+        var singleCurrencyTotal = new Amount(0, currency);
+        foreach (var item in orderPart.LineItems)
+        {
+            singleCurrencyTotal += item.LinePrice;
+        }
```
DAud-IcI, line 185: This can be simplified by passing `orderPart.LineItems.Sum()` to the constructor.

src/Modules/OrchardCore.Commerce/Abstractions/IStripePaymentService.cs
```diff
@@ -25,6 +26,16 @@ public interface IStripePaymentService
     /// </summary>
     Task<PaymentIntent> GetPaymentIntentAsync(string paymentIntentId);
 
+    /// <summary>
+    /// Calculates payment amount based on the given <paramref name="defaultTotalValue"/> and <paramref name="currencyType"/>.
+    /// </summary>
+    long GetPaymentAmount(decimal defaultTotalValue, string currencyType);
```
DAud-IcI, line 32: I don't like these argument names, because they describe how the method is currently used but without knowing that context they are pretty mysterious.  Why not just pass in one `Amount total` argument? I don't think there is any possible scenario where these two shouldn't be tied together. Also you are already using `Amount` for the other method.

src/Modules/OrchardCore.Commerce/Controllers/PaymentController.cs
```diff
@@ -150,6 +157,59 @@ public async Task<IActionResult> Validate()
...
+        {
+            OrderId = order.ContentItemId,
+            PaymentIntentId = paymentIntent.Id,
+        });
+
+        return View(new
+        {
+            SingleCurrencyTotal = singleCurrencyTotal,
+            NetTotal = singleCurrencyTotal,
+            GrossTotal = new Amount(0, currency),
```
DAud-IcI, line 206: This feels like an unnecessary workaround. You could just fix the line 38 in StripeCheckout.cshtml so it still works if the GrossTotal prop is not passed instead of passing a dummy value.

src/Modules/OrchardCore.Commerce/Services/StripePaymentService.cs
```diff
@@ -277,16 +253,38 @@ public async Task UpdateOrderToPaymentFailedAsync(PaymentIntent paymentIntent)
...
+    {
+        long amountForPayment;
+
+        // If I convert it to conditional expression, it will warn me to extract it again.
+#pragma warning disable IDE0045 // Convert to conditional expression
+        // We need to convert the value (decimal) to long.
+        // https://stripe.com/docs/currencies#zero-decimal
+        if (CurrencyCollectionConstants.ZeroDecimalCurrencies.Contains(currencyType))
+        {
+            amountForPayment = (long)Math.Round(defaultTotalValue);
```
DAud-IcI, line 266: Instead of defining a temporary variable you could return the value directly. (I think if you do that you won't have that IDE0045 warning either.)

src/Modules/OrchardCore.Commerce/Services/StripePaymentService.cs
```diff
@@ -333,6 +331,54 @@ public async Task<IEnumerable<OrderLineItem>> CreateOrderLineItemsAsync(Shopping
...
+    private static void AlterOrder(
+        ContentItem order,
+        PaymentIntent paymentIntent,
+        Amount defaultTotal,
+        ShoppingCartViewModel cartViewModel,
+        IEnumerable<OrderLineItem> lineItems)
+    {
+        order.Alter<OrderPart>(orderPart =>
+        {
+            orderPart.Charges.Clear();
```
DAud-IcI, line 343: Can Charges be non-empty at this point?

src/Modules/OrchardCore.Commerce/Views/OrderPart.cshtml
```diff
@@ -45,6 +47,18 @@
                 </div>
             </div>
         </div>
+        @if (string.Equals(Model.OrderPart.Status.Text, OrderStatuses.Pending, StringComparison.OrdinalIgnoreCase))
+        {
+            <div class="my-2 d-flex justify-content-center">
+                <div class="pay-button-container">
+                    <a href="@Href($"~/PaymentRequest/{Model.OrderPart.ContentItem.ContentItemId}")" class="pay-button btn btn-success justify-content-center">
```
DAud-IcI, line 54: Please use `@Orchard.Action<PaymentController>(controller => controller.PaymentRequest(Model.OrderPart.ContentItem.ContentItemId))` instead of hard coding the route.

src/Modules/OrchardCore.Commerce/Views/Payment/PaymentRequest.cshtml
```diff
@@ -0,0 +1,46 @@
...
+    <input type="hidden" id="OrderPart_ShippingAddress_Address_PostalCode" value="@shippingAddress.PostalCode" />
+    <input type="hidden" id="OrderPart_ShippingAddress_Address_Province" value="@shippingAddress.Province" />
+
+    <input type="hidden" id="OrderPart_BillingAddress_Address_Name" value="@billingAddress.Name" />
+    <input type="hidden" id="OrderPart_BillingAddress_Address_City" value="@billingAddress.City" />
+    <input type="hidden" id="OrderPart_BillingAddress_Address_Region" value="@billingAddress.Region" />
+    <input type="hidden" id="OrderPart_BillingAddress_Address_StreetAddress1" value="@billingAddress.StreetAddress1" />
+    <input type="hidden" id="OrderPart_BillingAddress_Address_StreetAddress2" value="@billingAddress.StreetAddress2" />
+    <input type="hidden" id="OrderPart_BillingAddress_Address_PostalCode" value="@billingAddress.PostalCode" />
+    <input type="hidden" id="OrderPart_BillingAddress_Address_Province" value="@billingAddress.Province" />
```
DAud-IcI, line 41: Can you display the two `AddressField` editors instead of hard coding the fields? This will not look good if someone overrides their AddressField_Edit shape and then expect the changed template to carry over to this page.

test/OrchardCore.Commerce.Tests.UI/Tests/OrderTests/BehaviorOrderTests.cs
```diff
@@ -141,14 +134,75 @@ public BehaviorOrderTests(ITestOutputHelper testOutputHelper)
...
             browser);
 
-    private static Task AddNewProductAsync(UITestContext context) => context.ClickReliablyOnAsync(By.Id("addButton"));
+    [Theory, Chrome]
+    public Task OrderPaymentRequestButtonShouldWorkProperly(Browser browser) =>
+        ExecuteTestAfterSetupAsync(
+            async context =>
+            {
+                await context.SignInDirectlyAndGoToDashboardAsync();
+                await context.CreateNewContentItemAsync(Order);
```
DAud-IcI, line 150: Could you create a pending order as sample recipe content? It would be good for demoing as well as make testing here simpler.

DAud-IcI (changes requested):

src/Modules/OrchardCore.Commerce/Controllers/PaymentController.cs
porgabi: It seems the method can be entirely replaced with a `Sum()` call: `orderTotals = orderPart.LineItems.Select(item => item.LinePrice).Sum();`

DAud-IcI: Yeah, but you still pass an `Amount` instance to `CreatePaymentIntentAsync` so you might as well stick with creating it here.

porgabi: Whoops, I was looking at the wrong thing. `StripePaymentService`'s `CalculateOrderTotals()` is what I was referring to as something that can be removed. I see what you mean for this one, will update.

src/Modules/OrchardCore.Commerce/Services/StripePaymentService.cs
porgabi: This was code that I just moved into a method, but I'll update it accordingly.

src/Modules/OrchardCore.Commerce/Services/StripePaymentService.cs
porgabi: I think it's possible when an Order was paid for once and then set to Pending again.

DAud-IcI: Sounds like fraud but that's none of my business...

src/Modules/OrchardCore.Commerce/Views/Payment/PaymentRequest.cshtml
porgabi: Currently these are hidden fields, so overriding the aforementioned shape shouldn't affect them.

Would it be preferable to make the fields editable the same way as during a regular checkout? Now that I think about it, that would make sense, I originally added the hidden fields simply because the values were needed for the Stripe payment form.

src/Modules/OrchardCore.Commerce/Views/OrderPart.cshtml
porgabi: This returns `/OrchardCore.Commerce/Payment/PaymentRequest`, which is not quite what is necessary (`/checkout/paymentrequest/4kx5ap5ds7qvs3swwtn15m13md`), and it looks like the path cannot be customized to such an extent using this method. Are there any other viable alternatives?

DAud-IcI: Very strange. Use the `asp-*` argument tag helpers then, like
```
                    <a asp-area="OrchardCore.Commerce"
                       asp-controller="@typeof(PaymentController).ControllerName()"
                       asp-action="@nameof(PaymentController.PaymentRequest)"
                       asp-route-orderId="@Model.OrderPart.ContentItem.ContentItemId"
```

src/Modules/OrchardCore.Commerce/Controllers/PaymentController.cs
```diff
@@ -157,40 +159,42 @@ public async Task<IActionResult> Validate()
...
 
-        // If status is not Pending or there are no line items, there is nothing to be done.
-        if (!string.Equals(orderPart.Status.Text, OrderStatuses.Pending, StringComparison.OrdinalIgnoreCase) ||
-            !orderPart.LineItems.Any())
+        // If there are no line items, there is nothing to be done.
+        if (!orderPart.LineItems.Any())
         {
-            return NotFound();
+            await _notifier.InformationAsync(H["This Order contains no line items, so there is nothing to be paid."]);
+            return LocalRedirect($"~/Contents/ContentItems/{orderId}");
```
DAud-IcI, line 178: Use the `this.RedirectToContentDisplay(orderId)` extension method everywhere. It's both cleaner and returns the correct URL even if autoroute part or similar is used.

src/Modules/OrchardCore.Commerce/Controllers/PaymentController.cs
```diff
@@ -157,40 +159,42 @@ public async Task<IActionResult> Validate()
...
+        // If status is not Pending, there is nothing to be done.
+        if (!string.Equals(orderPart.Status.Text, OrderStatuses.Pending, StringComparison.OrdinalIgnoreCase))
         {
-            singleCurrencyTotal += item.LinePrice;
+            await _notifier.InformationAsync(H["This Order is no longer pending."]);
+            return LocalRedirect($"~/Contents/ContentItems/{orderId}");
         }
 
+        var currency = orderPart.LineItems[0].LinePrice.Currency;
+        var singleCurrencyTotal = new Amount(orderPart.LineItems.Select(item => item.LinePrice).Sum().Value, currency);
```
DAud-IcI, line 189: You were right about using `orderPart.LineItems.Select(item => item.LinePrice).Sum()` to get the sum. So we won't actually need the separate `currency` variable:
```suggestion
        var singleCurrencyTotal = orderPart.LineItems.Select(item => item.LinePrice).Sum();
```

src/Modules/OrchardCore.Commerce/Controllers/PaymentController.cs
```diff
@@ -157,40 +159,42 @@ public async Task<IActionResult> Validate()
...
         if (singleCurrencyTotal.Value <= 0)
         {
-            return NotFound();
+            await _notifier.InformationAsync(H["This Order's line items have no cost, so there is nothing to be paid."]);
+            return LocalRedirect($"~/Contents/ContentItems/{orderId}");
         }
 
         var stripeApiSettings = (await _siteService.GetSiteSettingsAsync()).As<StripeApiSettings>();
-        var paymentAmount = _stripePaymentService.GetPaymentAmount(singleCurrencyTotal.Value, currency.CurrencyIsoCode);
+        var paymentAmount = _stripePaymentService.GetPaymentAmount(new Amount(singleCurrencyTotal.Value, currency));
```
DAud-IcI, line 197: It's not necessary to clone `singleCurrencyTotal`. Because `currency` is the same as `singleCurrencyTotal.Currency`, so you can just call `_stripePaymentService.GetPaymentAmount(singleCurrencyTotal);`

src/Modules/OrchardCore.Commerce/Services/StripePaymentService.cs
DAud-IcI: Thanks!

PR #359, opened 2023-09-28
OCC-191: Add widget to the setup recipe to access the user address and user details pages

### LINKED ISSUE #350: User informations (OCC-191)
Users should be able to edit their informations (addresses, details) and consult their orders from the Front end.

[Jira issue](https://lombiq.atlassian.net/browse/OCC-191)

DAud-IcI:
[OCC-191](https://lombiq.atlassian.net/browse/OCC-191)
Fixes #350

### DIFF
--- a/test/OrchardCore.Commerce.Tests.UI/Tests/UserTests/UserPersistenceTests.cs
+++ b/test/OrchardCore.Commerce.Tests.UI/Tests/UserTests/UserPersistenceTests.cs
@@ -41,8 +41,9 @@ public class UserPersistenceTests : UITestBase
         ExecuteTestAfterSetupAsync(
             async context =>
             {
-                await context.SignInDirectlyAsync();
-                await context.GoToRelativeUrlAsync("/user/addresses");
+                context.Missing(By.ClassName("user-addresses-widget"));
+                await context.SignInDirectlyAndGoToHomepageAsync();
+                await context.ClickReliablyOnAsync(By.ClassName("user-addresses-widget"));
 
                 await context.ClickAndFillInWithRetriesAsync(
                     By.Id("UserAddressesPart_ShippingAddress_Address_Name"),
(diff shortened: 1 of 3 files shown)


test/OrchardCore.Commerce.Tests.UI/Tests/UserTests/UserPersistenceTests.cs
```diff
@@ -41,8 +41,9 @@ public UserPersistenceTests(ITestOutputHelper testOutputHelper)
         ExecuteTestAfterSetupAsync(
             async context =>
             {
-                await context.SignInDirectlyAsync();
-                await context.GoToRelativeUrlAsync("/user/addresses");
+                context.Missing(By.ClassName("user-addresses-widget"));
```
DemeSzabolcs, line 44: Shouldn't we also check the other "widget" (`user-details-widget`) the same way? (I know it's technically one).

Here: https://github.com/OrchardCMS/OrchardCore.Commerce/blob/5f11e9b4e3e0da889d7c9f48ee999230b7227397/test/OrchardCore.Commerce.Tests.UI/Tests/UserTests/UserPersistenceTests.cs#L127-L128

By checking if the icon is missing if we are logged out, and clicking on it etc.

DemeSzabolcs (changes requested):

test/OrchardCore.Commerce.Tests.UI/Tests/UserTests/UserPersistenceTests.cs
DAud-IcI: I don't think there is a point to it, because it's the same template. Even checking the `context.Missing()` here felt kind of unnecessary, because we shouldn't have to test stock OC features such as the `Authenticated` layer working correctly.

DemeSzabolcs: Alright then!

Issue #365, opened 2023-10-11
Boolean product attributes not stored in the order screen. (OCC-195)


DAud-IcI:
### Describe the bug

The Boolean Product Attribute Field is kind of broken, see symptoms below.

### To Reproduce

1. Edit the "Product" content type (or create a new content type with ProductPart and PricePart).
2. Click _Add Field_ and select _Boolean Product Attribute Field_.
3. Order any Product (e.g. https://localhost:5001/testproduct) without ticking the new checkbox.
4. **In the cart there is no attribute indicator below the product name.**
5. Purchase succeeds, but **the Attributes column is empty** both in Order display and editor.
6. Order the same Product again, but this time check the box.
7. On the cart screen it doesn't show the product
8. Trying to click the Pay button **results in this async error** notification:
```
Microsoft.CSharp.RuntimeBinder.RuntimeBinderException: 'OrchardCore.Commerce.ProductAttributeValues.BooleanProductAttributeValue' does not contain a definition for 'PredefinedValue' at CallSite.Target(Closure , CallSite , Object ) at System.Dynamic.UpdateDelegates.UpdateAndExecute1[T0,TRet](CallSite site, T0 arg0) at OrchardCore.Commerce.Models.ShoppingCartItem.<>c.<CreateOrderLineFromShoppingCartItemAsync>b__22_1(IProductAttributeValue value) in E:\Repositories\orchard-core\OrchardCore.Commerce\src\Modules\OrchardCore.Commerce\Models\ShoppingCartItem.cs:line 127 
```

### Expected behavior

- The attribute should be applied even if its value is false.
- When displaying attributes in the cart or the order view/editor it should display the Boolean attribute in the `{fieldLabel ?? fieldName}: {Yes/No}` format. For example: `Deluxe Version: Yes` or `Insurance: No`
- Including a boolean attribute shouldn't during checkout.
- Boolean attributes should be displayed and editable in the order editor.

If it helps, this attribute type should be considered as if it was predefined with `true` and `false` possible values. It may benefit to have a container value derived from `IPredefinedValuesProductAttributeValue<bool>` similar to `TextProductAttributeValue` for text attributes.

[Jira issue](https://lombiq.atlassian.net/browse/OCC-195)

porgabi:
> Boolean attributes should be displayed and editable in the order editor.

I assume it'd make sense to handle numeric attributes the same way too, so I'll add that in this one as well. Let me know if I shouldn't, though.

PR #368, opened 2023-10-11
OCC-196: Product Attributes should display the field label or name instead in cart and order display

### LINKED ISSUE #366: Product Attributes should display the field label or name instead in cart and order display (OCC-196)
### Describe the bug

When you display a product attribute in the cart and in the order, for some reason it shows the content type before the colon instead of the field label. The order editor is not affected by this.

### To Reproduce

1. Order a small Test Price Variant Product (https://localhost:5001/Contents/ContentItems/testpricevariantproduct000)
2. See the Product column on the Your Cart screen. It says:
```
Test Price Variant Product
PriceVariantsProduct: Small
```
3. Similarly, if you complete the order on the Success screen `PriceVariantsProduct: Small` is displayed in the Attributes column.

### Expected behavior

In both cases it should've displyed `Size: Small`.

### Screenshots

image

[Jira issue](https://lombiq.atlassian.net/browse/OCC-196)

porgabi:
[OCC-196](https://lombiq.atlassian.net/browse/OCC-196)

Fixes #366

### DIFF
--- a/src/Modules/OrchardCore.Commerce/Views/BooleanProductAttributeField.cshtml
+++ b/src/Modules/OrchardCore.Commerce/Views/BooleanProductAttributeField.cshtml
@@ -1,5 +1,6 @@
 @using OrchardCore.Commerce.Models
 @using OrchardCore.Commerce.Settings
+
 @{
     int index = Model.Index;
     var attr = Model.AttributeDescription as ProductAttributeDescription;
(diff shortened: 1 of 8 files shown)


src/Modules/OrchardCore.Commerce/Views/BooleanProductAttributeField.cshtml
```diff
@@ -8,9 +9,9 @@
 }
 
 <input type="hidden" name="line.Attributes[@index].Key" value="@attr.PartName.@attr.Name" />
-<input type="hidden" name="line.Attributes[@index].Value" value="false" />
```
porgabi, line 11: Seems like a duplicate of the input directly below. It was causing invalid HTML (`error: Duplicate form control name "line.Attributes[1].Value" (form-dup-name)`) and reoving it doesn't seem to cause additional issues based on my testing.

DAud-IcI: This is the equivalent of how `@Html.CheckBoxFor()` works by default so the form value is sent even for false. But as pointed out [here](https://github.com/OrchardCMS/OrchardCore.Commerce/issues/365#:~:text=The%20attribute%20should%20be%20applied%20even%20if%20its%20value%20is%20false.) it doesn't work so it's fine to remove.

DAud-IcI (approved):

PR #377, opened 2023-11-11
OCC-199: Prevent checking out if there are unavailable products in the cart

### LINKED ISSUE #375: Prevent checking out if there are unavailable products in the cart (OCC-199)
### Describe the bug

Currently you can try to check out an order with "Inventory: 0" and it will only fail after the payment is already made.

### To Reproduce

1. Edit a product by setting its Inventory to 0.
2. Try to check it out.
3. See `InvalidOperationException` _after_ the payment has been processed!

### Expected behavior

- The checkout button on the "Your Cart" screen should be greyed out when there is such a product in the cart.
- An error notification of the inventory shortage should be displayed on the cart screen.
- The checkout screen should redirect you back to the cart so this problem can't be bypassed by manually navigating to the `~/checkout` url.

### Screenshots

gif

[Jira issue](https://lombiq.atlassian.net/browse/OCC-199)

porgabi:
[OCC-199](https://lombiq.atlassian.net/browse/OCC-199)
Fixes #375


src/Modules/OrchardCore.Commerce/Startup.cs
```diff
@@ -170,6 +170,9 @@ public override void ConfigureServices(IServiceCollection services)
         // Page
         services.AddScoped<IDataMigration, PageMigrations>();
 
+        // Checkout
+        services.AddScoped<ICheckoutEvents, InventoryCheckoutEvents>();
```
sarahelsaig, line 174: We only need inventory events if the `OrchardCore.Commerce.Inventory` feature is enabled. So please move this into `InventoryStartup` class below.

test/OrchardCore.Commerce.Tests.UI/Tests/InventoryTests/InventoryBehaviourTests.cs
```diff
@@ -73,6 +75,56 @@ public InventoryBehaviourTests(ITestOutputHelper testOutputHelper)
...
+                await context.GoToAdminRelativeUrlAsync($"/Contents/ContentItems/{TestProduct}/Edit");
+                await context.ClickReliablyOnAsync(By.XPath("//label[contains(., 'Allows Back Order')]"));
+                await context.ClickAndFillInWithRetriesAsync(By.Name("InventoryPart.Inventory[TESTPRODUCT]"), "0");
+                await context.ClickPublishAsync();
+
+                // Verify checkout is unavailable.
+                await context.GoToCartAsync();
+                context.Exists(By.XPath($"//p[contains(., '{CheckoutUnavailableMessage}')]"));
+                await context.GoToCheckoutAsync();
+                context.Exists(By.XPath($"//div[contains(., '{CheckoutUnavailableMessage}')]"));
```
sarahelsaig, line 96: Avoid just checking div text, because there are lots of divs on the page. Check for the class first like
`context.Exists(By.XPath($"//div[contains(@class, 'message-error') and contains(., '{CheckoutUnavailableMessage}')]"));`

sarahelsaig, line 96: If you replace `Exists`/`Missing` with `CheckExistence` then this can be turned into a reusable method.

sarahelsaig:
I found a way to still reproduce the original error. If you set the inventory of Test Free Product (which allows back-order) to 0 and try to check out you still get the "Inventory value cannot be negative." error.

sarahelsaig (changes requested):

src/Modules/OrchardCore.Commerce.Payment/Services/PaymentService.cs
```diff
@@ -130,6 +130,12 @@ public class PaymentService : IPaymentService
             CheckoutShapes = checkoutShapes,
         };
 
+        // Checkout should not be possible if any of the items are unpurchasable.
```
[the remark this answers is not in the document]
porgabi: It's a [word](https://www.merriam-webster.com/dictionary/unpurchasable) (and even if it wasn't, I would have invented it, it just makes sense for it to exist!).

src/Modules/OrchardCore.Commerce/Services/ShoppingCartHelpers.cs
```diff
@@ -120,13 +122,35 @@ public class ShoppingCartHelpers : IShoppingCartHelpers
             totals = lines.CalculateTotals().ToList();
         }
 
-        // The values are rounded to avoid storing more precision than what the currency supports.
+        // Checkout should not be possible if any of the items are unpurchasable.
```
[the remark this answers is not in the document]
porgabi: The model we are dealing with here is `ShoppingCartViewModel`, while inside `ViewModelCreatedAsync()` it's the `CheckoutViewModel` that's handled (after making the above requested change to pass in `ICheckoutViewModel`).

I suppose both could be passed in as nullable and then the method can handle them based on which one is not null. Though there might be a more elegant way to do this?

```
    public Task ViewModelCreatedAsync(
        IList<ShoppingCartLineViewModel> lines,
        ICheckoutViewModel checkoutViewModel = null,
        ShoppingCartViewModel shoppingCartViewModel = null)
```

src/Modules/OrchardCore.Commerce/Views/ShoppingCart/Index.cshtml
```diff
@@ -6,11 +10,17 @@
...
-
-    <div class="col-xs-12 col-sm-1">
-        <a asp-action="Index"
-           asp-controller="Payment"
-           asp-area="OrchardCore.Commerce.Payment"
-           class="btn btn-primary checkout">@T["Checkout"]</a>
-    </div>
+    @if (Model.IsInvalid)
+    {
+        <p>@T["Checkout unavailable — an item is out of stock."]</p>
```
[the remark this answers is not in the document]
porgabi: Would it be worthwhile adding the same to `CheckoutViewModel` too?

src/Modules/OrchardCore.Commerce.Payment/Services/PaymentService.cs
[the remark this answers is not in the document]
sarahelsaig: image
I assume this word was green lit by the Redundancy Department of Redundancy.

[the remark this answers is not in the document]
porgabi: I might be having a smooth-brain moment, but isn't it the same as with, for example, available/unavailable/not available?

porgabi:
> I found a way to still reproduce the original error. If you set the inventory of Test Free Product (which allows back-order) to 0 and try to check out you still get the "Inventory value cannot be negative." error.

Looks like the allows back order value is not properly taken into account during checkout. How should this work when the inventory still has items available and back ordering is allowed -- inventory should be decreased as normal, right? As long as it doesn't go below 0, that is.

src/Modules/OrchardCore.Commerce/Services/ShoppingCartHelpers.cs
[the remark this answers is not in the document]
sarahelsaig: OCC is so complex, we really can't afford code duplication of hard assumptions about which features are enabled. So it's important to treat features like Inventory as plugins instead of wiring them into the main program. In this case you could do the following:
1. Move the business logic from InventoryCheckoutEvents into a service (maybe `ProductInventoryService`?) using your original method signature, i.e. `Task<bool> VerifyLinesAsync(IList<ShoppingCartLineViewModel> lines)`.
2. Add a `ViewModelCreatedAsync` method to `IShoppingCartEvents` as well.
3. Now both `InventoryCheckoutEvents.ViewModelCreatedAsync()` and `InventoryShoppingCartEvents.ViewModelCreatedAsync()` can call the service with the shared logic and update their respective view-models without harming future extension possibilities.

src/Modules/OrchardCore.Commerce/Events/InventoryCheckoutEvents.cs
```diff
@@ -39,10 +48,28 @@ public Task<bool> ViewModelCreatedAsync(IList<ShoppingCartLineViewModel> lines)
...
+                {
+                    shoppingCartViewModel.InvalidReasons.Add(H["{0} is out of stock.", inventoryIdentifier]);
+                }
+
+                return Task.CompletedTask;
             }
         }
 
-        return Task.FromResult(cannotCheckout);
+        if (checkoutViewModel != null)
```
sarahelsaig, line 64: Why is this necessary? I think if in the future we'll have other events that can invalidate the cart, then we shouldn't erase that here.

src/Modules/OrchardCore.Commerce/Services/ShoppingCartHelpers.cs
```diff
@@ -122,40 +121,22 @@ public class ShoppingCartHelpers : IShoppingCartHelpers
...
 
-        model.IsInvalid = cannotCheckout;
-
+        var model = new ShoppingCartViewModel { Id = cart.Id };
         model.Totals.AddRange(totals.Any() ? totals.Round() : new List<Amount> { new(0, lines[0].LinePrice.Currency) });
-
         model.Headers.AddRange(headers);
         model.Lines.AddRange(lines);
 
+        // Checkout should not be possible if any of the items are unpurchasable.
```
sarahelsaig, line 136: Since we no longer set the `model.IsInvalid` on the outside I don't think this comment is needed. This event may be used for other reasons just as well.

sarahelsaig, line 136: Since we no longer set the `model.IsInvalid` on the outside I don't think this comment is needed. This event may be used for other reasons besides validation.

src/Modules/OrchardCore.Commerce/Views/ShoppingCart/Index.cshtml
```diff
@@ -10,9 +10,15 @@
 
 <section id="shopping-cart">
     <shape type="ShoppingCartTable" prop-ShoppingCartViewModel="@Model"></shape>
-    @if (Model.IsInvalid)
+    @if (Model.InvalidReasons.Any())
     {
-        <p>@T["Checkout unavailable — an item is out of stock."]</p>
+        <div class="checkout-unavailable">@T["Checkout unavailable due to the following:"]</div>
```
sarahelsaig, line 15:
```suggestion
        <div class="checkout-unavailable">@T["Checkout unavailable due to the following reasons:"]</div>
```
I think you need that extra word here? At least I remember seeing that specific phrasing in the past.

sarahelsaig (changes requested):

sarahelsaig:
> > I found a way to still reproduce the original error. If you set the inventory of Test Free Product (which allows back-order) to 0 and try to check out you still get the "Inventory value cannot be negative." error.
> 
> Looks like the allows back order value is not properly taken into account during checkout. How should this work when the inventory still has items available and back ordering is allowed -- inventory should be decreased as normal, right? As long as it doesn't go below 0, that is.

Yeah, inventory should still be decreased. I think you just have to change [this line](https://github.com/OrchardCMS/OrchardCore.Commerce/blob/6b89d9d2151b44b3fe9557d3038164941d5d2313/src/Modules/OrchardCore.Commerce/Services/LocalInventoryProvider.cs#L85) to also take `inventoryPart.AllowsBackOrder` into consideration

src/Modules/OrchardCore.Commerce/Services/LocalInventoryProvider.cs
```diff
@@ -81,7 +81,10 @@ private async Task UpdateInventoryAsync(ProductPart productPart, int difference,
             var relevantInventory = inventoryPart.Inventory.FirstOrDefault(entry =>
                 entry.Key == inventoryIdentifier || entry.Key == inventoryRootIdentifier);
 
-            var newValue = relevantInventory.Value + difference;
+            var newValue = relevantInventory.Value + difference < 0 && inventoryPart.AllowsBackOrder.Value
+                ? 0
+                : relevantInventory.Value + difference;
```
sarahelsaig, line 86: Why did you clamp it to zero? I thought we agreed that the inventory should still be decreased. Doesn't this lose the back-order information?

src/Modules/OrchardCore.Commerce/Services/ProductInventoryService.cs
```diff
@@ -25,4 +34,33 @@ public async Task<IList<ShoppingCartItem>> UpdateInventoriesAsync(IList<Shopping
...
+                !inventoryPart.AllowsBackOrder.Value &&
+                !inventoryPart.IgnoreInventory.Value;
+
+            if (cannotCheckout)
+            {
+                return Task.FromResult(true);
+            }
+        }
+
+        return Task.FromResult(cannotCheckout);
```
sarahelsaig, line 64: This will always be false, so it would be better for readability to spell it out here.
(then the `var cannotCheckout` can be moved inside the loop)

src/Modules/OrchardCore.Commerce/Abstractions/IProductInventoryService.cs
```diff
@@ -14,4 +15,9 @@ public interface IProductInventoryService
     /// </summary>
     /// <param name="items">The quantities and products whose inventories need to be adjusted.</param>
     Task<IList<ShoppingCartItem>> UpdateInventoriesAsync(IList<ShoppingCartItem> items);
+
+    /// <summary>
+    /// Verifies the inventory state of the provided <paramref name="lines"/>.
+    /// </summary>
```
sarahelsaig, line 21: You still didn't explain what the return value does and it's not quite obvious.
```suggestion
    /// <summary>
    /// Verifies the inventory state of the provided <paramref name="lines"/>.
    /// </summary>
    /// <returns><see langword="true"/> if the line items can't be checked out.</returns>
```

src/Modules/OrchardCore.Commerce/Events/InventoryShoppingCartEvents.cs
```diff
@@ -13,17 +14,31 @@ public class InventoryShoppingCartEvents : ShoppingCartEventsBase
...
+        _productInventoryService = productInventoryService;
         H = localizer;
     }
 
     public override int Order => 0;
 
+    public override async Task ViewModelCreatedAsync(ShoppingCartViewModel viewModel)
+    {
+        var cannotCheckout = await _productInventoryService.VerifyLinesAsync(viewModel.Lines);
+        if (cannotCheckout)
```
sarahelsaig, line 34: Why not merge them?
```suggestion
        if (await _productInventoryService.VerifyLinesAsync(viewModel.Lines))
```

src/Modules/OrchardCore.Commerce/Events/InventoryShoppingCartEvents.cs
```diff
@@ -13,17 +14,31 @@ public class InventoryShoppingCartEvents : ShoppingCartEventsBase
...
+        {
+            viewModel.InvalidReasons.Add(H["An item in cart is out of stock."]);
+        }
+
+        return;
```
sarahelsaig, line 39: This return doesn't really do anything.

src/Modules/OrchardCore.Commerce/Events/InventoryCheckoutEvents.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    private readonly IProductInventoryService _productInventoryService;
+
+    public InventoryCheckoutEvents(
+        IProductInventoryService productInventoryService) => _productInventoryService = productInventoryService;
+
+    public async Task ViewModelCreatedAsync(IList<ShoppingCartLineViewModel> lines, ICheckoutViewModel checkoutViewModel)
+    {
+        var cannotCheckout = await _productInventoryService.VerifyLinesAsync(lines);
+        if (cannotCheckout)
```
sarahelsaig, line 19: These could be merged too, the "cannot checkout" information is communicated by the `IsInvalid = true`.
```suggestion
        if (await _productInventoryService.VerifyLinesAsync(lines))
```

src/Modules/OrchardCore.Commerce/Events/InventoryCheckoutEvents.cs
```diff
@@ -0,0 +1,26 @@
...
+        {
+            checkoutViewModel.IsInvalid = true;
+        }
+
+        return;
```
sarahelsaig, line 24: This is unnecessary as well.

sarahelsaig (changes requested):

src/Modules/OrchardCore.Commerce/Services/LocalInventoryProvider.cs
porgabi: Isn't a negative value invalid? In the previous discussion I was thinking of a scenario where, for example, inventory is 5 and a customer orders 3 items (I was unsure if that should decrease inventory or just count as back-ordered items).

So the goal with this change was to make it so if the order amount would decrease the inventory below 0, it should be set to 0 instead. I wasn't aware that this had to keep track of the amount of back-ordered items, it never worked that way, but then I assume it should just keep the negative value?

sarahelsaig: [Back-ordering](https://www.investopedia.com/terms/b/backorder.asp#:~:text=A%20backorder%20is%20an%20order%20for%20a%20good%20or%20service%20that%20cannot%20be%20filled%20at%20the%20current%20time%20due%20to%20a%20lack%20of%20available%20supply.) is when you order more than what's on inventory, not that you what's left in stock. If that setting is allowed, then a negative inventory means that the merchant has to order more from their supplier to resolve the inventory deficits.

So like I said, just change [this line](https://github.com/OrchardCMS/OrchardCore.Commerce/blob/6b89d9d2151b44b3fe9557d3038164941d5d2313/src/Modules/OrchardCore.Commerce/Services/LocalInventoryProvider.cs#L85) to also check `inventoryPart.AllowsBackOrder`.

porgabi: Makes sense, I was operating under the assumption that the back-order information is not meant to be stored this way. I'll amend.

PR #399, opened 2024-01-13
OCC-208: Checkout exception if the Billing/ShippingAddressFieldEditor shapes are not present

### LINKED ISSUE #391: Checkout exception if the Billing/ShippingAddressFieldEditor shapes are not present (OCC-208)
### Describe the bug

If the two (or any of the two) address shapes aren't rendered, `AddressFieldDisplayDriver`'s `UpdateAsync()` is still called during the checkout process, which ends up throwing an exception as the fields are null. This is also the case when the address fields' values are set directly.

image

### To Reproduce

1. Override the Checkout shape with a version that doesn't include the two address shapes.
2. Set the required values of the address fields directly.
3. Try to check out.

### Expected behavior

If these shapes aren't present and the requires values are set directly, the update method should not be called.

[Jira issue](https://lombiq.atlassian.net/browse/OCC-208)

porgabi:
[OCC-208](https://lombiq.atlassian.net/browse/OCC-208)

Fixes #391


src/Modules/OrchardCore.Commerce.ContentFields/Drivers/AddressFieldDisplayDriver.cs
```diff
@@ -84,7 +84,13 @@ bool IsRequiredFieldEmpty(string value, string key)
             return true;
         }
 
-        if (!await updater.TryUpdateModelAsync(viewModel, Prefix) ||
+        var updated = await updater.TryUpdateModelAsync(viewModel, Prefix);
```
hishamco, line 87:
```suggestion
        var updated = await updater.TryUpdateModelAsync(viewModel, Prefix);

```

src/Modules/OrchardCore.Commerce.Payment/Services/PaymentService.cs
```diff
@@ -198,6 +198,21 @@ await _paymentProvidersLazy
...
+            {
+                if (orderPart.BillingAddress.Address.Name is null && orderPart.ShippingAddress.Address.Name is not null)
+                {
+                    orderPart.BillingAddress = orderPart.ShippingAddress;
+                }
+
+                if (orderPart.ShippingAddress.Address.Name is null && orderPart.BillingAddress.Address.Name is not null)
+                {
+                    orderPart.ShippingAddress = orderPart.BillingAddress;
+                }
```
hishamco, line 213:
```suggestion
                billingAddress ??= shippingAddress;
		shippingAddress ??= billingAddress;
```

hishamco (changes requested):

src/Modules/OrchardCore.Commerce.Payment/Services/PaymentService.cs
sarahelsaig: The `BillingAddress` and `ShippingAddress` properties [are never null](https://github.com/OrchardCMS/OrchardCore.Commerce/blob/18a900038099da680ea49c99c4ab63732564235f/src/Libraries/OrchardCore.Commerce.Abstractions/Models/OrderPart.cs#L42-L43). Also their `Address` properties (e.g. `OrderPart.BillingAddress.Address`) [are always initialized as well](https://github.com/OrchardCMS/OrchardCore.Commerce/blob/18a900038099da680ea49c99c4ab63732564235f/src/Libraries/OrchardCore.Commerce.Abstractions/Fields/AddressField.cs#L8). That's why it's necessary to drill down and individually check the name field (which is a known required field) instead of your suggestion.

@porgabi On the other hand I don't `not null` checks of each condition are necessary. It only matters if both addresses are empty, but then you would get the same exception as before anyway. So it's cleaner and probably slightly faster to remove them.

src/Modules/OrchardCore.Commerce.Payment/Services/PaymentService.cs
```diff
@@ -198,6 +198,21 @@ await _paymentProvidersLazy
             orderPart.LineItems.SetItems(lineItems);
             orderPart.Status.Text = OrderStatuses.Pending.HtmlClassify();
 
+            // When billing and shipping addresses should match, if an address field is null but the other one is not,
+            // fill out the null field's data.
```
sarahelsaig, line 202: Could you rephrase this comment to describe the situation where this is necessary? Currently it feels more like it's just describing the code below, which is not that useful.

sarahelsaig (changes requested):

PR #420, opened 2024-04-09
Integrate exactly payments for EU/UK merchants

sarahelsaig:


src/Modules/OrchardCore.Commerce.Payment.Exactly/Views/ExactlySettings.Edit.cshtml
```diff
@@ -0,0 +1,55 @@
+@using Microsoft.AspNetCore.Html
+@using Microsoft.AspNetCore.Mvc.Localization
+@using OrchardCore
+@using OrchardCore.Commerce.Payment.Exactly.Controllers
+@model OrchardCore.Commerce.Payment.Exactly.Models.ExactlySettings
+
+<p class="alert alert-info">
+    @T["Please be sure to use <a href=\"{0}\">this link</a> to create your exactly account.", "https://application.exactly.com/?utm_source=partner&utm_medium=kirill&utm_campaign=LOMBIQ"]
+    @T["That way <a href=\"https://lombiq.com/\">Lombiq</a> (the steward of the Orchard Core Commerce project) will get a commission on the payment fees, which helps cover some of the development costs of OCC."]
+    @T["This is at no cost to you; the fees you pay are the same either way."]
```
Piedone, line 10: Add this to the docs too.

OrchardCore.Commerce.sln
DemeSzabolcs, line 1: I guess this (Strip not configured and it's not working):
image

is related to this: https://github.com/OrchardCMS/OrchardCore.Commerce/issues/424
Right?

src/Modules/OrchardCore.Commerce.Payment.Exactly/Views/ExactlySettings.Edit.cshtml
```diff
@@ -0,0 +1,55 @@
...
+</p>
+
+<div class="mb-3">
+    <div>
+        <a id="ExactlySettings__testButton"
+            href="@(Orchard.Action<ExactlyController>(controller => controller.VerifyApi()))"
+           class="btn btn-sm btn-primary">
+            @T["Verify currently saved API configuration"]
```
DemeSzabolcs, line 18: Not in this PR/issue, but something similar would be nice for Stripe too.

DemeSzabolcs, line 18: After the verification is successful, a content item is created: 
image

But it throws an error if we want to view it, since I believe the product sku is null for the test item: 
image

Also, if we want to keep this content item, then the title of it should reflect that it's created from the verification and it's not a real one.

OrchardCore.Commerce.sln
DemeSzabolcs, line 1: I tested Exactly with multiple currencies. EUR and USD worked fine, but GBP didn't (unsupported currency). However I didn't find a documentation on the supported currencies, so this might be an expected behavior.

DemeSzabolcs (changes requested):

src/Modules/OrchardCore.Commerce/Migrations/OrderMigrations.cs
```diff
@@ -280,4 +281,31 @@ public int UpdateFrom6()
...
+                            new ListValueOption { Name = Arrived, Value = OrderStatusCodes.Arrived },
+                            new ListValueOption { Name = PaymentFailed, Value = OrderStatusCodes.PaymentFailed },
+                        ],
+                        DefaultValue = OrderStatusCodes.Pending,
+                        Editor = EditorOption.Radio,
+                    }))
+            );
+
+        return 8;
+    }
```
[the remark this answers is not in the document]
sarahelsaig: I have added the PaymentFailed that was missing, even though it's used by Stripe.

docs/features/exactly-payment.md
```diff
@@ -0,0 +1,34 @@
...
+There are available test cards that can be found in [Stripe's documentation](https://stripe.com/docs/testing).
+
+There are multiple test cards that can simulate any scenario, including error codes. Here are two examples:
+
+| Brand      | Number              | CVC          | Date            | Result                                           |
+|------------|---------------------|--------------|-----------------|--------------------------------------------------|
+| Visa       | 4000 0000 0000 7775 | Any 3 digits | Any future date | the card has insufficient funds                  |
+| Visa       | 4000 0000 0000 3220 | Any 3 digits | Any future date | success during 3DS Auth (3DS is always expected) |
+| Visa       | 4000 0084 0000 1280 | Any 3 digits | Any future date | the card fails 3DS Auth (3DS is always expected) |
+| Mastercard | 5555 5555 5555 4444 | Any 3 digits | Any future date | Mastercard test card                             |
```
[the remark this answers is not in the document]
sarahelsaig: The text above the table was accidentally pasted, I just wanted to copy the table as a formatting template. Now that I double check it, some of the card numbers are indeed the same but not all.

src/Modules/OrchardCore.Commerce.Payment.Exactly/Models/ChargeResponse.cs
```diff
@@ -0,0 +1,152 @@
...
+    public ChargeAttributes Attributes { get; set; }
+
+    [SuppressMessage(
+        "StyleCop.CSharp.NamingRules",
+        "SA1313:Parameter names should begin with lower-case letter",
+        Justification = "Necessary for localization extractor.")]
+    public static IDictionary<string, LocalizedString> GetResultCodes(IStringLocalizer<ChargeResponse> T) =>
+        new Dictionary<string, LocalizedString>
+        {
+            ["success"] = T["processing of the transaction was successfully completed"],
```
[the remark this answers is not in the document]
sarahelsaig: They are copied [from the documentation](https://exactly.com/docs/api#tag/Transactions/operation/getTransactionById) as-is. I don't really want to edit them because that will make future maintenance harder.
image
Also they are only used like `"Your transaction has failed: {0}."` (e.g. "Your transaction has failed: transaction failed, no specific details could be provided.") so it works.

src/Modules/OrchardCore.Commerce.Payment.Exactly/Models/ChargeRequest.cs
```diff
@@ -0,0 +1,88 @@
...
+    public string ProjectId { get; set; }
+    public string PaymentMethod { get; set; } = "card";
+    public string Amount { get; set; }
+    public string Currency { get; set; }
+    public string ReferenceId { get; set; }
+    public string CustomerDescription { get; set; }
+    public string ReturnUrl { get; set; }
+    public string CustomerId { get; set; }
+    public string Email { get; set; }
+    public int Lifetime { get; set; } = 3600;
```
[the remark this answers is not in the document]
sarahelsaig: It's serialized and sent by the request. I don't think Refit automatically converts between TimeSpan and int so we'd have to add another technical property that does the conversion. Doesn't feel like it's worth the effort until we find a need to change from the default value.

src/Modules/OrchardCore.Commerce.Payment.Exactly/Views/CheckoutExactly.cshtml
```diff
@@ -0,0 +1,74 @@
...
+                    }
+
+                    await sleep(200);
+                }
+            }
+
+            setButtons(true);
+        });
+    });
+</script>
```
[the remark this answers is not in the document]
sarahelsaig: This is much simpler, and adding Lombiq.NodeJs.Extensions dependency for a single JS file felt like overkill.

OrchardCore.Commerce.sln
sarahelsaig: Yes.

src/Modules/OrchardCore.Commerce.Payment.Exactly/Views/ExactlySettings.Edit.cshtml
sarahelsaig: Ok https://github.com/OrchardCMS/OrchardCore.Commerce/issues/426

OrchardCore.Commerce.sln
sarahelsaig: I don't know, I've only heard that Netherlands is not supported ([see](https://lombiq.atlassian.net/browse/OCC-203?focusedCommentId=84364)).

src/Modules/OrchardCore.Commerce.Payment.Exactly/Views/CheckoutExactly.cshtml
Piedone, line 1: Doesn't this need a CSP config?

src/Modules/OrchardCore.Commerce.Payment.Exactly/Views/ExactlySettings.Edit.cshtml
Piedone, line 10: Don't divide localized strings along English boundaries, because the order of sentences might not make sense in other languages. Have the whole thing a single string (that'll look really bad in the code, but this can't be properly localized).

src/Modules/OrchardCore.Commerce.Payment.Exactly/Views/CheckoutExactly.cshtml
sarahelsaig: Nope. It's nothing fancy, just a basic script block with no dependencies.

Piedone: How's the payment happening, like where are you providing your credit card details? Does it operate fully with redirects?

sarahelsaig: We don't handle credit card details. Only redirects, as we have discussed. Here is an overview:
- JS script sends a POST request that only contains the contents of the checkout page (i.e. addresses).
- C# backend creates a new Order content item from the checkout data and the stored shopping cart. 
- C# backend sends a POST request to the Exactly API including the order total and the return URL.
- JS script gets a redirect URL (on Exactly's domain) as response, then navigates there.
- Payment continues on Exactly, then redirects back to the return URL.
- C# backend validates the transaction state, updates Order and redirects to Success page.

Piedone: Great, thanks. I suggest copying this to the Readme of the module too.

OrchardCore.Commerce.sln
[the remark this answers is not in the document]
sarahelsaig: I can validate the signup notification, but I don't think there is any point to test the fields. They are simple string backed forms, without using external resources they can contain basically anything. Also wouldn't want to include our keys in the test even if they are sandbox only.

sarahelsaig: They said only EUR and USD are supported in sandbox. I've added a not in the doc [here](https://github.com/OrchardCMS/OrchardCore.Commerce/pull/420/commits/983f551bb38e345f614a633aac275a11abc5e6cd).
Apparently they only enable specific currencies for each live project based on agreement with the merchant. Weird.

DemeSzabolcs (approved):

PR #421, opened 2024-04-10
OCC-229: Inventory part with non-restricted product attributes can't be added to the cart

### LINKED ISSUE #419: Inventory part with non-restricted product attributes can't be added to the cart (OCC-229)
### Describe the bug

It's specifically only broken when the product has attributes but none are restricted to predefined values.

### To Reproduce

1. Create a new  content type.
2. Add Product part, Price part and Inventory part.
3. Add a text product attribute field (name doesn't matter).
4. Create a content item, make sure to set the inventory count to a positive number.
5. View the content item.
6. Observe that:
   - ✔️ The inventory number is as expected.
   - 🛑 The Quantity field defaults to 0 instead of 1.
7. Set the Quantity to 1 and click Checkout.
8. Observe that:
   - 🛑 An error notification saying  "There are not enough X left in stock." is displayed.
   - 🛑 The product is not actually added to the cart.

### Expected behavior

It looks like the inventory shopping cart event can't resolve the entry in the inventory dictionary so it defaults to assuming the inventory is 0. This should not be the case.

### Screenshots

[Jira issue](https://lombiq.atlassian.net/browse/OCC-229)

sarahelsaig:
[OCC-229](https://lombiq.atlassian.net/browse/OCC-229)
Fixes #419

### DIFF
--- a/src/Modules/OrchardCore.Commerce.Inventory/Drivers/InventoryPartDisplayDriver.cs
+++ b/src/Modules/OrchardCore.Commerce.Inventory/Drivers/InventoryPartDisplayDriver.cs
@@ -1,3 +1,4 @@
+using GraphQL.Introspection;
 using Microsoft.AspNetCore.Http;
 using OrchardCore.Commerce.Inventory.Models;
 using OrchardCore.Commerce.Inventory.ViewModels;
@@ -5,6 +6,7 @@ using OrchardCore.ContentManagement.Display.ContentDisplay;
 using OrchardCore.ContentManagement.Display.Models;
 using OrchardCore.DisplayManagement.ModelBinding;
 using OrchardCore.DisplayManagement.Views;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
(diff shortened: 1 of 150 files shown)


src/Modules/OrchardCore.Commerce.Inventory/Drivers/InventoryPartDisplayDriver.cs
```diff
@@ -36,7 +38,7 @@ public class InventoryPartDisplayDriver : ContentPartDisplayDriver<InventoryPart
             var currentSku = _hca.HttpContext?.Request.Form["ProductPart.Sku"].ToString().ToUpperInvariant();
             var skuBefore = viewModel.Inventory.FirstOrDefault().Key != null
                 ? viewModel.Inventory.FirstOrDefault().Key.Split('-')[0]
-                : "DEFAULT";
+                : currentSku;
```
DemeSzabolcs, line 41: Shouldn't `DefaultKey` should be used here? (Or the const should be removed) Because the const now is not used anywhere.

src/Modules/OrchardCore.Commerce/Drivers/ProductPartDisplayDriver.cs
DemeSzabolcs, line 1: I reset the webapp, then tried to repro the problem with your changes (so there shouldn't be any repro): https://github.com/OrchardCMS/OrchardCore.Commerce/issues/419#issue-2232395838

I did these:
> Create a new content type.
Add Product part, Price part and Inventory part.
Add a text product attribute field (name doesn't matter).

image

But when I try to create a new item, I don't have SKU or Quantity: 
image

Also, I get this in the log: 
```
OrchardCore.ContentManagement.Display.ContentItemDisplayCoordinator|ERROR|IContentPartDisplayDriver thrown from OrchardCore.Commerce.Drivers.ProductPartDisplayDriver by ArgumentNullException System.ArgumentNullException: Value cannot be null. (Parameter 'key')
   at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
   at System.Collections.Generic.Dictionary`2.Add(TKey key, TValue value)
   at OrchardCore.Commerce.Drivers.ProductPartDisplayDriver.BuildViewModelAsync(ProductPartViewModel viewModel, ProductPart part) in G:\Work\OrchardCore.Commerce\src\Modules\OrchardCore.Commerce\Drivers\ProductPartDisplayDriver.cs:line 114
   at OrchardCore.Commerce.Drivers.ProductPartDisplayDriver.<>c__DisplayClass4_0.<<Edit>b__0>d.MoveNext() in G:\Work\OrchardCore.Commerce\src\Modules\OrchardCore.Commerce\Drivers\ProductPartDisplayDriver.cs:line 37
```

DemeSzabolcs (changes requested):

src/Modules/OrchardCore.Commerce/Drivers/ProductPartDisplayDriver.cs
DemeSzabolcs, line 1: I found another bug. When creating a content item for the first time, if you trying to save the item, without the SKU, you will get an error message (that's okay), but the SKU field disappears: 
image

DemeSzabolcs (changes requested):

src/Modules/OrchardCore.Commerce/Drivers/ProductPartDisplayDriver.cs
sarahelsaig: That sounded suspiciously unrelated, so I checked the current `main` branch and the same is true there. Not to expand the scope here, I will open a separate issue.

DemeSzabolcs: Alright!

DemeSzabolcs (approved):

PR #435, opened 2024-04-30
Update OC 1.8.3

agriffard:
https://github.com/OrchardCMS/OrchardCore/releases/tag/v1.8.3


Directory.Packages.props
```diff
@@ -14,26 +14,26 @@
...
-    <PackageVersion Include="OrchardCore.Indexing.Abstractions" Version="1.8.2" />
-    <PackageVersion Include="OrchardCore.Localization" Version="1.8.2" />
-    <PackageVersion Include="OrchardCore.Logging.NLog" Version="1.8.2" />
-    <PackageVersion Include="OrchardCore.Module.Targets" Version="1.8.2" />
-    <PackageVersion Include="OrchardCore.Navigation.Core" Version="1.8.2" />
-    <PackageVersion Include="OrchardCore.Settings" Version="1.8.2" />
-    <PackageVersion Include="OrchardCore.Templates" Version="1.8.2" />
-    <PackageVersion Include="OrchardCore.Title" Version="1.8.2" />
-    <PackageVersion Include="OrchardCore.Workflows.Abstractions" Version="1.8.2" />
+    <PackageVersion Include="OrchardCore.Application.Cms.Targets" Version="1.8.3" />
```
Piedone, line 17: I think the OCC packages should always target the .0 minor version of the latest version; i.e., 1.8.0 in this case, so people can use the packages regardless of them being on 1.8.0, 1.8.1, 1.8.2, or 1.8.3 (currently they're forced to use at least 1.8.2). During development and testing though, we should use the very latest, if somehow this can be done. @sarahelsaig what do you think?

hishamco: Sometime patch version fixes few bug which is good to bump to the lastes

Piedone: Yes, but that's not what I'm talking about.

hishamco: Ah, got it now you was refering to OCC :)

sarahelsaig: Good idea, @Piedone!

At first I was going to suggest using [`VersionOverride`](https://learn.microsoft.com/en-us/nuget/consume-packages/central-package-management#overriding-package-versions), but on second thought it would be cleaner to define a variable in the `<PropertyGroup>` above like:
```xml
  <PropertyGroup>
    ...
    <OrchardCoreVersion Condition="'$(NuGetBuild)' == 'true'">1.8.0</OrchardCoreVersion>
    <OrchardCoreVersion Condition="'$(NuGetBuild)' != 'true'">1.8.3</OrchardCoreVersion>
  </PropertyGroup>
  <ItemGroup>
    ...
    <PackageVersion Include="OrchardCore.Application.Cms.Targets" Version="$(OrchardCoreVersion)" />
    <PackageVersion Include="OrchardCore.ContentFields" Version="$(OrchardCoreVersion)" />
    ...
  </ItemGroup>
```

The added bonus here is that going forward we will only have to change either one or two lines on a version upgrade instead of 16.

sarahelsaig: Turns out Lombiq.HelpfulLibraries requires OC 1.8.2 so for now replace 1.8.0 with 1.8.2 above.
https://github.com/Lombiq/Helpful-Libraries/issues/258

Also it looks like the Lombiq.* NuGet packages actually export the `NuGetBuild` symbol? This is very strange.
image

Piedone: That's because of https://github.com/Lombiq/GitHub-Actions/pull/343/files#diff-96e58ab979bf1e8b7d14178a048ebbe8901f627c6e3daf9abfcfe88d93f1282b. I'm fixing it.

PR #438, opened 2024-05-02
Docs requirements

agriffard:
Update docs requirements.

I updated the settings on Crowdin to not display this banner: 
image

On merge, it will trigger a new build and should update the docs taking this setting into account.


sarahelsaig:
Thank you! It works. That banner was really annoying.

PR #477, opened 2024-08-06
OCC-277: Get stripe confirm parameters async

### LINKED ISSUE #476: Abstract the Stripe GetConfirmPaymentParameters logic to service (OCC-277)

abstract the logic inside service

use new method of GetStripeConfirmParametersAsync

[Jira issue](https://lombiq.atlassian.net/browse/OCC-277)

infofromca:
Fixes #476

### DIFF
--- a/src/Modules/OrchardCore.Commerce.Payment.Stripe/Abstractions/IStripePaymentService.cs
+++ b/src/Modules/OrchardCore.Commerce.Payment.Stripe/Abstractions/IStripePaymentService.cs
@@ -65,6 +65,6 @@ public interface IStripePaymentService
     /// <summary>
     /// Get the confirmation parameters for Stripe.
     /// </summary>
-    /// <param name="middlewareUrl">The url for the middleware of Stripe.</param>
-    Task<PaymentIntentConfirmOptions> GetStripeConfirmParametersAsync(string middlewareUrl);
+    /// <param name="middlewareAbsoluteUrl">The url for the middleware of Stripe.</param>
+    Task<PaymentIntentConfirmOptions> GetStripeConfirmParametersAsync(string middlewareAbsoluteUrl);
 }
--- a/src/Modules/OrchardCore.Commerce.Payment.Stripe/Abstractions/IStripePaymentService.cs
+++ b/src/Modules/OrchardCore.Commerce.Payment.Stripe/Abstractions/IStripePaymentService.cs
@@ -61,4 +61,11 @@ public interface IStripePaymentService
         Controller controller,
         ContentItem order,
         PaymentIntent paymentIntent);
+
+    /// <summary>
+    /// Get the confirmation parameters for Stripe.
+    /// </summary>
+    /// <param name="middlewareUrl">The url for the middleware of Stripe.</param>
+    /// <returns><see cref="PaymentIntentConfirmOptions"/>.</returns>
+    Task<PaymentIntentConfirmOptions> GetStripeConfirmParametersAsync(string middlewareUrl);
 }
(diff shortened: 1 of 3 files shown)


src/Modules/OrchardCore.Commerce.Payment.Stripe/Services/StripePaymentService.cs
```diff
@@ -215,9 +216,51 @@ public async Task<PaymentIntent> CreatePaymentIntentAsync(Amount total)
...
+            Line1 = address.StreetAddress1 ?? string.Empty,
+            Line2 = address.StreetAddress2 ?? string.Empty,
+            PostalCode = address.PostalCode ?? string.Empty,
+            State = address.Province ?? string.Empty,
+        };
     private async Task<PaymentIntent> GetOrUpdatePaymentIntentAsync(
-        string paymentIntentId,
-        Amount defaultTotal)
+    string paymentIntentId,
+    Amount defaultTotal)
```
sarahelsaig, line 263: Revert this change, parameters should be indented from the starting line of the function signature.

src/Modules/OrchardCore.Commerce.Payment.Stripe/Abstractions/IStripePaymentService.cs
```diff
@@ -61,4 +61,11 @@ public interface IStripePaymentService
         Controller controller,
         ContentItem order,
         PaymentIntent paymentIntent);
+
+    /// <summary>
+    /// Get the confirmation parameters for Stripe.
+    /// </summary>
+    /// <param name="middlewareUrl">The url for the middleware of Stripe.</param>
+    /// <returns><see cref="PaymentIntentConfirmOptions"/>.</returns>
```
sarahelsaig, line 69:
```suggestion
```
This return text adds no useful information, so it can be removed.

src/Modules/OrchardCore.Commerce.Payment.Stripe/Abstractions/IStripePaymentService.cs
```diff
@@ -61,4 +61,11 @@ public interface IStripePaymentService
...
+    Task<PaymentIntentConfirmOptions> GetStripeConfirmParametersAsync(string middlewareUrl);
```
sarahelsaig, line 70: Shouldn't the parameter type be `Uri` instead?

sarahelsaig (changes requested):
Thanks, looks good. I just need a few small tweaks before it can be merged.

src/Modules/OrchardCore.Commerce.Payment.Stripe/Abstractions/IStripePaymentService.cs
infofromca: not necessary in this case, and not increase complicated

sarahelsaig: After looking the code over again, this has to be absolute URL. Normally in ASP.Net Core you pass around relative virtual URLs, like `~/checkout/middleware/Stripe`. Which wouldn't work in this case because the full URL is sent out to a remote server.
Please rename the parameter to `middlewareAbsoluteUrl`

sarahelsaig (changes requested):

src/Modules/OrchardCore.Commerce.Payment.Stripe/Abstractions/IStripePaymentService.cs
infofromca: renamed to middlewareAbsoluteUrl

PR #485, opened 2024-08-23
OFFI-99: Inferred parameters are not always recognized correctly causing exceptions

wAsnk:
[OFFI-99](https://lombiq.atlassian.net/browse/OFFI-99)

### DIFF
--- a/src/Modules/OrchardCore.Commerce/Endpoints/Api/PaymentEndpoint.cs
+++ b/src/Modules/OrchardCore.Commerce/Endpoints/Api/PaymentEndpoint.cs
@@ -23,10 +23,10 @@ public static class PaymentEndpoint
 
     [Authorize(AuthenticationSchemes = "Api")]
     private static async Task<IResult> AddFreeAsync(
-         string? shoppingCartId,
-         IAuthorizationService authorizationService,
-         HttpContext httpContext,
-         IPaymentService paymentService
+         [FromRoute] string? shoppingCartId,
+         [FromServices] IAuthorizationService authorizationService,
+         [FromServices] HttpContext httpContext,
+         [FromServices] IPaymentService paymentService
        )
     {
         if (!await authorizationService.AuthorizeAsync(httpContext.User, ApiPermissions.CommerceApi))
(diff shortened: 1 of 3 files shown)


src/Modules/OrchardCore.Commerce/Endpoints/Api/PaymentEndpoint.cs
```diff
@@ -23,10 +23,10 @@ public static IEndpointRouteBuilder AddFreeEndpoint(this IEndpointRouteBuilder b
 
     [Authorize(AuthenticationSchemes = "Api")]
     private static async Task<IResult> AddFreeAsync(
-         string? shoppingCartId,
-         IAuthorizationService authorizationService,
-         HttpContext httpContext,
-         IPaymentService paymentService
+         [FromRoute] string? shoppingCartId,
```
hishamco, line 26: Why not use them as properties decorated with `FromRoute`, `FromService` instead of injecting them within the constructor?

sarahelsaig: This is a static class. How would that work?

hishamco: You can use it like this

```csharp
[HttpGet("api/products/{id}")]
public IActionResult GetProduct([FromRoute] int id)
{
    // id will be populated with the value from the route
}
```

hishamco: I remember I used [FromRoute] in the RazorPage model property

```csharp
[FromRoute("id")]
public string Id { get; set; }
```

wAsnk: Could you tell me why this is a better approach? What's the benefit? @hishamco

hishamco: Much cleaner and no need to use constructor injection

sarahelsaig: I think this could only be done by turning this from a minimal API endpoint class into an API controller? But as I recall OC is moving the other direction (away from API controllers) for performance reasons.

PR #493, opened 2024-09-04
Add more fields to parts by using partial class

### LINKED ISSUE #492: How can I add more fields to part , such as ProductPart (OCC-291)
Taks ProductPart as example:
I need 1--public MediaField ProductImage { get; set; }
2--public MediaField SecondaryProductImage { get; set; }
3--public VideoField PrimaryProductVideo { get; set; }
public partial class ProductPart

[Jira issue](https://lombiq.atlassian.net/browse/OCC-291)

infofromca:
Fixes #492

### DIFF
--- a/src/Libraries/OrchardCore.Commerce.Abstractions/Models/OrderPart.cs
+++ b/src/Libraries/OrchardCore.Commerce.Abstractions/Models/OrderPart.cs
@@ -10,7 +10,9 @@ using System.Collections.Generic;
 
 namespace OrchardCore.Commerce.Abstractions.Models;
 
-public class OrderPart : ContentPart
+#pragma warning disable S2333 // Redundant modifiers should not be used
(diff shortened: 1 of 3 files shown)


sarahelsaig:
This will not work, because [all partial-type definitions meant to be parts of the same type must be defined in the same assembly and the same module (.exe or .dll file). Partial definitions can't span multiple modules](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/partial-classes-and-methods#:~:text=All%20partial%2Dtype%20definitions%20meant%20to%20be%20parts%20of%20the%20same%20type%20must%20be%20defined%20in%20the%20same%20assembly%20and%20the%20same%20module%20(.exe%20or%20.dll%20file).%20Partial%20definitions%20can't%20span%20multiple%20modules.).

PR #494, opened 2024-09-11
Update Orchard Core 2.0.0

hishamco:
@sarahelsaig This might need to bump all of the Lombiq packages to use `OC 2.0.0`, please let me know if you need help there or merge after updating Lombiq packages


sarahelsaig:
It's not so simple, we already have [a separate branch](https://github.com/OrchardCMS/OrchardCore.Commerce/compare/main...task/system-text-json-migration) with a ton of fixes to address the many breaking changes in OC 2.0. 
image

Also because of the Lombiq libraries this issue is blocker: https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/issues/709
Once it's resolved and the new NuGet packages released (some time later this month) we will update the `task/system-text-json-migration` branch and merge into `main` from that. So this PR won't be necessary but thanks anyway.

PR #502, opened 2024-09-29
OCC-194: Upgrade to Orchard Core 2.0 once it's released and remove Newtonsoft.Json from the code base

### LINKED ISSUE #362: Upgrade to Orchard Core 2.0 once it's released and remove Newtonsoft.Json from the code base (OCC-194)
Orchard Core 2.0 will finally migrate to System.Text.Json and [drop Newtonsoft.Json support](https://github.com/OrchardCMS/OrchardCore/blob/main/src/docs/releases/1.9.0.md#drop-newtonsoftjson-support). This much anticipated event should be followed up in Orchard Core Commerce as well.

Blocking:
- OC 2.0 release (couldn't find a release issue in the OC repo)
- https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/issues/709

----

Obsolete description:

Currently OCC has serialization via Newtonsoft Json.NET and System.Text.Json. Both are actually used in different places. This makes things unnecessarily complicated as serializers have to be implemented in two different frameworks.

Orchard Core still uses Json.NET at the core of its content system ([see](https://github.com/OrchardCMS/OrchardCore/blob/main/src/OrchardCore/OrchardCore.ContentManagement.Abstractions/ContentElement.cs#L31)) and so do many other third party modules and libraries that build on OC. While the OC team has expressed intent to eventually switch over to STJ, this is not in the near future. Additionally based on previous comments (for example [here](https://github.com/OrchardCMS/OrchardCore.Commerce/issues/53#issuecomment-586467524)) we might want to rewrite the STJ implementation from scratch if this ever happens.
In the meantime, we should do the following to make the code more maintainable and approachable:
- Remove the STJ converters
- If possible, rename the Json.NET converters without the "Legacy" prefix. (if OC ever changes over to STJ we will delete these converters anyway so I don't expect that we should ever have two different sets.
- Partially rewrite `ShoppingCartSerializer`, `IProductAttributeProvider` and any other file where `using System.Text.Json;` is found to use Json.NET instead.

Alternatively we may fully commit to STJ-only, but I think at this point this would be premature, an enormous technical debt and a source of future headaches.

[Jira issue](https://lombiq.atlassian.net/browse/OCC-194)

Psichorex:
[OCC-194](https://lombiq.atlassian.net/browse/OCC-194)
Fixes #362


src/OrchardCore.Commerce.Web/OrchardCore.Commerce.Web.csproj
```diff
@@ -17,9 +17,10 @@
     <PackageReference Include="Lombiq.Tests.UI.Shortcuts" />
 
     <!-- The Orchard Core packages must override the version to target the latest patch version in this project. See the
-         Directory.Packages.props file for more details. -->
-    <PackageReference Include="OrchardCore.Application.Cms.Targets" VersionOverride="1.8.3" />
-    <PackageReference Include="OrchardCore.Logging.NLog" VersionOverride="1.8.3" />
+         Directory.Packages.props file for more details. In branches using OC preview, the value should be
+         "$(OrchardCoreVersion)" instead, which is the same as removing the VersionOverride but more maintainable. -->
+    <PackageReference Include="OrchardCore.Application.Cms.Targets" VersionOverride="$(OrchardCoreVersion)" />
+    <PackageReference Include="OrchardCore.Logging.NLog" VersionOverride="$(OrchardCoreVersion)" />
```
sarahelsaig, line 23: It's good that `OrchardCoreVersion` is only 2.0.0 (as it should always be the x.y.0 version), but the Web.csproj should use the latest. Please revert the comment and set the `VersionOverride` to hard coded [2.0.2](https://www.nuget.org/packages/OrchardCore/2.0.2).

sarahelsaig (changes requested):

src/OrchardCore.Commerce.Web/OrchardCore.Commerce.Web.csproj
```diff
@@ -16,10 +16,8 @@
     <PackageReference Include="Lombiq.Tests.UI.AppExtensions" />
     <PackageReference Include="Lombiq.Tests.UI.Shortcuts" />
 
-    <!-- The Orchard Core packages must override the version to target the latest patch version in this project. See the
-         Directory.Packages.props file for more details. -->
-    <PackageReference Include="OrchardCore.Application.Cms.Targets" VersionOverride="1.8.3" />
-    <PackageReference Include="OrchardCore.Logging.NLog" VersionOverride="1.8.3" />
+    <PackageReference Include="OrchardCore.Application.Cms.Targets" VersionOverride="2.0.2" />
```
sarahelsaig, line 19:
```suggestion
    <!-- The Orchard Core packages must override the version to target the latest patch version in this project. See the
         Directory.Packages.props file for more details. -->
    <PackageReference Include="OrchardCore.Application.Cms.Targets" VersionOverride="2.0.2" />
```

PR #512, opened 2024-12-12
NEST-537: Upgrading to OC 2.1.0

BenedekFarkas:
[NEST-537](https://lombiq.atlassian.net/browse/NEST-537)

PR #514, opened 2025-01-22
OFFI-200: Update OC and Lombiq versions.

sarahelsaig:
[OFFI-200](https://lombiq.atlassian.net/browse/OFFI-200)

### DIFF
--- a/test/OrchardCore.Commerce.Tests.UI/Tests/BasicTests/SecurityScanningTests.cs
+++ b/test/OrchardCore.Commerce.Tests.UI/Tests/BasicTests/SecurityScanningTests.cs
@@ -1,5 +1,4 @@
 using Lombiq.Tests.UI.SecurityScanning;
-using Lombiq.Tests.UI.Shortcuts.Middlewares;
 using Xunit;
 using Xunit.Abstractions;
 
--- a/test/OrchardCore.Commerce.Tests.UI/Tests/BasicTests/SecurityScanningTests.cs
+++ b/test/OrchardCore.Commerce.Tests.UI/Tests/BasicTests/SecurityScanningTests.cs
@@ -40,14 +40,7 @@ public class SecurityScanningTests : UITestBase
                         @".*/\?.*products\..*");
                 },
                 maxActiveScanDurationInMinutes: 5,
-                maxRuleDurationInMinutes: 1,
-                additionalPermittedErrorLinePatterns:
-                [
-                    "System.IO.DirectoryNotFoundException: Could not find a part of the path",
-                    "System.IO.IOException: The filename, directory name, or volume label syntax is incorrect",
-                    "System.InvalidOperationException: This action intentionally causes an exception!",
-                    nameof(ExceptionContextLoggingMiddleware),
-                ]));
+                maxRuleDurationInMinutes: 1));
 
     private static void FalsePositive(
         SecurityScanConfiguration configuration,
--- a/test/OrchardCore.Commerce.Tests.UI/Tests/BasicTests/SecurityScanningTests.cs
+++ b/test/OrchardCore.Commerce.Tests.UI/Tests/BasicTests/SecurityScanningTests.cs
@@ -17,9 +17,6 @@ public class SecurityScanningTests : UITestBase
             context => context.RunAndConfigureAndAssertFullSecurityScanForContinuousIntegrationAsync(
                 configuration =>
                 {
-                    // Not relevant for OrchardCore.Commerce, sites using it should implement their own custom error pages.
-                    configuration.DontScanErrorPage = true;
-
                     configuration.DisableActiveScanRule(
                         6,
                         "Path Traversal (all paths are virtual so it's not a real concern, also creates too many errors)");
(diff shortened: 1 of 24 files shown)


test/OrchardCore.Commerce.Tests.UI/Tests/BasicTests/SecurityScanningTests.cs
```diff
@@ -41,19 +40,14 @@ public Task FullSecurityScanShouldPass() =>
...
-                        !message.ContainsOrdinalIgnoreCase("System.InvalidOperationException: This action intentionally causes an exception!"));
-
-                logsWithoutUnwantedExceptionMessages.ShouldNotContain(item => item.Contains("|ERROR|"));
-            });
+                maxRuleDurationInMinutes: 1,
+                additionalPermittedErrorLinePatterns:
+                [
+                    "System.IO.DirectoryNotFoundException: Could not find a part of the path",
+                    "System.IO.IOException: The filename, directory name, or volume label syntax is incorrect",
+                    "System.InvalidOperationException: This action intentionally causes an exception!",
```
BenedekFarkas, line 48: These are covered by using
```csharp
changeConfiguration: configuration => configuration.UseAssertAppLogsForSecurityScan()
```
instead of
```
configuration => configuration.AssertAppLogsAsync = async webApplicationInstance =>
...
```

test/OrchardCore.Commerce.Tests.UI/Tests/BasicTests/SecurityScanningTests.cs
```diff
@@ -41,19 +40,14 @@ public Task FullSecurityScanShouldPass() =>
...
+                    nameof(ExceptionContextLoggingMiddleware),
```
BenedekFarkas, line 49: Doesn't this cause _every_ error to be ignored?

test/OrchardCore.Commerce.Tests.UI/Tests/BasicTests/SecurityScanningTests.cs
sarahelsaig: `RunAndConfigureAndAssertFullSecurityScanForContinuousIntegrationAsync` [already calls `UseAssertAppLogsForSecurityScan` in the background](https://github.com/Lombiq/UI-Testing-Toolbox/blob/5c2b55c49bf319ba2d822ba4eaaa98b0e6208372/Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs#L74) with these passed as additional permitted error lines. The first two are indeed not necessary, but the "This action intentionally causes an exception" is [not part of the default set](https://github.com/Lombiq/UI-Testing-Toolbox/blob/dev/Lombiq.Tests.UI/SecurityScanning/OrchardCoreUITestExecutorConfigurationExtensions.cs#L36-L61). But I'm not sure why. It should be added to the defaults in UITT.

BenedekFarkas: > RunAndConfigureAndAssertFullSecurityScanForContinuousIntegrationAsync [already calls UseAssertAppLogsForSecurityScan in the background](https://github.com/Lombiq/UI-Testing-Toolbox/blob/5c2b55c49bf319ba2d822ba4eaaa98b0e6208372/Lombiq.Tests.UI/SecurityScanning/SecurityScanningUITestContextExtensions.cs#L74) with these passed as additional permitted error lines. The first two are indeed not necessary

I know, that's why I suggested to use it, because it simplifies the test configuration. I did that already in my PR: https://github.com/OrchardCMS/OrchardCore.Commerce/pull/512/files#diff-147599d8d8983cbbed8db8009ed9dfa10e86f0cf7471aac8d39fb8d7de862ff1

> but the "This action intentionally causes an exception" is [not part of the default set](https://github.com/Lombiq/UI-Testing-Toolbox/blob/dev/Lombiq.Tests.UI/SecurityScanning/OrchardCoreUITestExecutorConfigurationExtensions.cs#L36-L61). But I'm not sure why. It should be added to the defaults in UITT.

Yes it does, that's the value behind [`ErrorController.ExceptionMessage`](https://github.com/Lombiq/UI-Testing-Toolbox/blob/v12.0.0/Lombiq.Tests.UI/SecurityScanning/OrchardCoreUITestExecutorConfigurationExtensions.cs#L57).

test/OrchardCore.Commerce.Tests.UI/Tests/BasicTests/SecurityScanningTests.cs
BenedekFarkas: At first glance that still seems too far reaching to me.

test/OrchardCore.Commerce.Tests.UI/Tests/BasicTests/SecurityScanningTests.cs
sarahelsaig: > I know, that's why I suggested to use it, because it simplifies the test configuration.

But this PR already wasn't setting `configuration.AssertAppLogsAsync`. Also like I said, because of `RunAndConfigureAndAssertFullSecurityScanForContinuousIntegrationAsync` in the test body [this line](https://github.com/OrchardCMS/OrchardCore.Commerce/pull/512/files#diff-147599d8d8983cbbed8db8009ed9dfa10e86f0cf7471aac8d39fb8d7de862ff1R43) in your PR is redundant.

test/OrchardCore.Commerce.Tests.UI/Tests/BasicTests/SecurityScanningTests.cs
sarahelsaig: `ExceptionContextLoggingMiddleware` detects when a request has thrown an exception and logs additional information about its HTTP context for easier diagnostics. It only uses `logger.LogError` for visibility: it's always next to a real error log anyway so, this simplifies the logger config.

If the real error is already suppressed, then not suppressing `ExceptionContextLoggingMiddleware` will cause a false error detection in `AssertAppLogsAsync`. If the real error is not suppressed, then we'll have a detection anyways and it doesn't matter. So logically this log should always be suppressed.

BenedekFarkas: Oh, OK, so it's just additional logging, but if there's a real error we shouldn't ignore, that won't have `ExceptionContextLoggingMiddleware` in its text.
I see you've got a UITT PR with this change, but I'll just add it directly in the UI test in my PR now, while I'm working on the rest of the DotNest upgrade.

test/OrchardCore.Commerce.Tests.UI/Tests/BasicTests/SecurityScanningTests.cs
BenedekFarkas: OIC, thanks!

wAsnk (approved):

PR #519, opened 2025-02-11
NEST-537: Upgrade web project to OC 2.1.6

BenedekFarkas:
[NEST-537](https://lombiq.atlassian.net/browse/NEST-537)


BenedekFarkas:
@sarahelsaig FYI

Piedone (approved):

Skrypt:
@sarahelsaig Why is it still targetting net8.0 and not net9.0 ?

sarahelsaig:
> Why is it still targetting net8.0 and not net9.0?

Projects maintained by Lombiq always target the current official LTS version of the .NET SDK (which is typically the even numbered version and [currently](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core) net8.0). I didn't give it much thought beyond following this practice, is there any particular reason why it should be .NET 9 instead?

sarahelsaig (approved):

Skrypt:
Yeah, OC is dual targetting that's why I asked. I expected to build this with net9.0 like OC.

sarahelsaig:
Good point, this may be a blind spot. I'm not opposed to it. Feel free to open an issue.

PR #523, opened 2025-02-20
More meaningful name and url

infofromca:


sarahelsaig:
Sorry for the late response. Could you explain how this change makes the name more meaningful? Seems like a matter of taste to me, which is not important for a tutorial example. You are supposed to replace those strings with the shop's actual name when you work on a real project anyway. (I'm closing this PR for now, but we can reopen if needed.)

PR #537, opened 2025-03-11
Fix launch settings for VS Code

Skrypt:
Under Windows, from what I tested, it fails starting the browser with VS Code with the current settings. This fix that issue.


sarahelsaig (approved):
Thanks!

PR #577, opened 2025-04-19
OCC-314: Remove node.js dependency

### LINKED ISSUE #527: Consider removing node.js dependency (OCC-314)

sarahelsaig:
[OCC-314](https://lombiq.atlassian.net/browse/OCC-314)
Fixes #527


src/Modules/OrchardCore.Commerce.ContentFields/OrchardCore.Commerce.ContentFields.csproj
```diff
@@ -44,15 +44,12 @@
...
-  <ItemGroup Condition="!Exists($(LombiqNodeJsExtensionsPath))">
-    <PackageReference Include="Lombiq.NodeJs.Extensions" />
-  </ItemGroup>
+  <Target Name="Copy JavaScript Assets"
+          AfterTargets="AfterResolveReferences">
+    <ItemGroup>
+      <JavaScriptAssets Include="Assets\Scripts\**\*.js"/>
+    </ItemGroup>
+    <Copy SourceFiles="@(JavaScriptAssets)" DestinationFolder="wwwroot\js\%(RecursiveDir)" SkipUnchangedFiles="true" />
+  </Target>
```
Piedone, line 53: Do we really need this? These files can just go directly into the wwwroot folder. For files coming from NPM packages such a copy (from _node_modules_ to _wwwroot_) would be interesting.

Also see the same copy targets in the other projects.

src/Modules/OrchardCore.Commerce/Assets/Styles/shopping-cart.css
Piedone, line 1: How about using the `&` nesting selector and changing the `class="shopping-cart-table-totals"` pattern to `class="shopping-cart-table totals"`? This is a breaking change, though, and needs to be documented.

src/Modules/OrchardCore.Commerce/Startup.cs
```diff
@@ -68,6 +69,7 @@ public override void ConfigureServices(IServiceCollection services)
         services.AddTagHelpers<MvcTitleTagHelper>();
         services.AddTransient<IConfigureOptions<ResourceManagementOptions>, ResourceManagementOptionsConfiguration>();
         services.AddScoped<IUserService, UserService>();
+        services.AddScoped<IResourceFilterProvider, ResourceFilters>();
```
Piedone, line 72: Let's have an extension method for this, since this is repeated across OSOCE too, like `AddResourceFilter<T>()`.

.github/workflows/build-and-test.yml
```diff
@@ -31,8 +31,15 @@ jobs:
     name: Codespell
     uses: Lombiq/GitHub-Actions/.github/workflows/codespell.yml@dev
 
+  lint:
+    name: Lint Script and Style
```
Piedone, line 35: We need MD linting too in this project.

Piedone (changes requested):

src/Modules/OrchardCore.Commerce.ContentFields/OrchardCore.Commerce.ContentFields.csproj
sarahelsaig: Ironically if I do that, then we have to bring back the package.json to manually configure the asset route to be in wwwroot for the linting. Also that won't be a problem here, but it wouldn't fly in projects where you also copy vendor files into wwwroot. Overall I think this is still cleaner and more intentional.

src/Modules/OrchardCore.Commerce/Assets/Styles/shopping-cart.css
sarahelsaig: Do we have to? I'm not comfortable making a breaking change just to save 3x20 characters.

Piedone: OK, let's leave it. It's just that the CSS isn't really nice this way, and I'd argue the classes would be better that way as well.

src/Modules/OrchardCore.Commerce.ContentFields/OrchardCore.Commerce.ContentFields.csproj
Piedone: There should be a better way. If we cut down on the needless complexity, then putting files into one folder just to copy them into another during build shouldn't be required.

What do you mean by "manually configure the asset route to be in wwwroot for the linting"? We already configure folders paths when invoking the workflow; can't that be utilized?

sarahelsaig: In the workflow we only provide the project paths, not the asset paths inside the project.

Piedone: Yes, but that can be changed. We can also let users configure the wwwroot, and recognize it by looking at the path.

.github/workflows/build-and-test.yml
sarahelsaig: Added, here is an example: https://github.com/OrchardCMS/OrchardCore.Commerce/actions/runs/14821047779/job/41608428919?pr=577#step:3:237

sarahelsaig: I've added MD linting, only allow it for the whole repo because giving more configurability here feels like an inconvenience and a bad idea. [Here is an example error.](https://github.com/OrchardCMS/OrchardCore.Commerce/actions/runs/14821047779/job/41608613936?pr=577#step:3:257)

image

src/Modules/OrchardCore.Commerce.ContentFields/wwwroot/js/Scripts/commerce-regions.js
Piedone, line 1: I guess this is an unwanted duplicate? And `src/Modules/OrchardCore.Commerce.Payment.Stripe/wwwroot/js/Scripts/stripe-payment-form.js`, `src/Modules/OrchardCore.Commerce.Payment/wwwroot/css/Styles/payment-form.css` too.

Piedone (changes requested):

src/Modules/OrchardCore.Commerce.ContentFields/wwwroot/js/Scripts/commerce-regions.js
sarahelsaig: No, they are just relocated. I have moved these files from _Assets_ to _wwwroot_ because you asked for it:

> [Do we really need this? These files can just go directly into the wwwroot folder. ](https://github.com/OrchardCMS/OrchardCore.Commerce/pull/577#discussion_r2059071705)

Piedone: But there are pairs with the same content for each of these files under the respective `wwwroot` folders.

{2A285BAF-668B-4464-9F06-5B6F936D3F99}

Piedone (approved):

Piedone (approved):

PR #602, opened 2025-06-09
Another SimpleCheckout for quick checkout processShow address

### LINKED ISSUE #517: SimpleCheckout for quick checkout process

The current checkout process requires users to fill in multiple details such as billing address, shipping address, and other fields, which can be time-consuming and may discourage quick purchases. For certain use cases, a simpler checkout process would improve user experience and conversion rates.  

I would like to request a **SimpleCheckout** option, where users only need to provide their **full name, email, and phone number** to complete the checkout process. This would be useful for digital products, local pickup orders, or cases where minimal information is required.  

- Allowing users to skip certain fields in the existing checkout process, but this still involves unnecessary steps.  
- Using guest checkout with auto-filled defaults, but it still requires address details.  
- Implementing a "Quick Buy" button that bypasses standard checkout but lacks customization.  

A dedicated **SimpleCheckout** option would streamline the process effectively while keeping the experience user-friendly.

infofromca:


infofromca:
@sarahelsaig  please review it. thanks

src/Libraries/OrchardCore.Commerce.Abstractions/Abstractions/ICheckoutViewModel.cs
```diff
@@ -14,9 +14,10 @@ public interface ICheckoutViewModel : IPaymentViewModel, IShape
 {
     string ShoppingCartId { get; }
     Amount GrossTotal { get; }
-    IEnumerable<SelectListItem> Regions { get; set; }
+    IEnumerable<Region> Regions { get; set; }
```
sarahelsaig, line 17: Why is this necessary? Such breaking changes should be avoided, because it will cause errors for anyone who customized their `Checkout` shape for any reason. I know of at least two such cases in client projects, so this not a speculative concern.

I suggest creating a separate new property if you need to reference the original region data as well. For example:

```csharp
    IEnumerable<SelectListItem> Regions { get; set; }
    IEnumerable<Region> RegionData { get; set; }
```

src/Modules/OrchardCore.Commerce.Payment/Services/CheckoutAddressService.cs
```diff
@@ -0,0 +1,23 @@
...
+    public CheckoutAddressService(ISiteService siteService) => _siteService = siteService;
+
+    public virtual async Task<bool> ShouldIgnoreAddressAsync(CheckoutViewModel checkoutViewModel)
+    {
+        var settings = (await _siteService.GetSiteSettingsAsync()).As<CheckoutAddressSettings>();
+        var shouldIgnore = settings.ShouldIgnoreAddress;
+
+        var total = checkoutViewModel.SingleCurrencyTotal.Value;
+
+        return shouldIgnore || total <= 0 || total > 10000;
```
sarahelsaig, line 21: Could you explain the `total > 10000`?

src/Modules/OrchardCore.Commerce.Payment/Services/CheckoutAddressService.cs
```diff
@@ -0,0 +1,23 @@
...
+    private readonly ISiteService _siteService;
+
+    public CheckoutAddressService(ISiteService siteService) => _siteService = siteService;
+
+    public virtual async Task<bool> ShouldIgnoreAddressAsync(CheckoutViewModel checkoutViewModel)
+    {
+        var settings = (await _siteService.GetSiteSettingsAsync()).As<CheckoutAddressSettings>();
+        var shouldIgnore = settings.ShouldIgnoreAddress;
+
+        var total = checkoutViewModel.SingleCurrencyTotal.Value;
```
sarahelsaig, line 19: I don't think that empty line is helpful here.
```suggestion
        var total = checkoutViewModel.SingleCurrencyTotal.Value;
```

src/Modules/OrchardCore.Commerce.Payment/Settings/CheckoutAddressSettingsDisplayDriver.cs
```diff
@@ -0,0 +1,66 @@
...
+    protected override string SettingsGroupId => GroupId;
+
+    public override async Task<IDisplayResult> EditAsync(ISite model, CheckoutAddressSettings section, BuildEditorContext context)
+    {
+        if (!await AuthorizeAsync())
+        {
+#pragma warning disable CS8603
+            return null;
+#pragma warning restore CS8603
+        }
```
sarahelsaig, line 42: Always include a comment explaining why you use pragma warning disable.
But in this case, I think you can just do this. (note the added `?` in the `Task<IDisplayResult?>` return type)

```csharp
    public override async Task<IDisplayResult?> EditAsync(ISite model, CheckoutAddressSettings section, BuildEditorContext context)
    {
        if (!await AuthorizeAsync()) return null;
```

src/Modules/OrchardCore.Commerce.Payment/Settings/CheckoutAddressSettingsDisplayDriver.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        context.AddTenantReloadWarningWrapper();
+
+        return Initialize<CheckoutAddressSettings>("CheckoutAddressSettings_Edit", model => model.ShouldIgnoreAddress = section.ShouldIgnoreAddress)
```
sarahelsaig, line 46:
```suggestion
        return Initialize<CheckoutAddressSettings>($"{nameof(CheckoutAddressSettings)}_Edit", model => model.ShouldIgnoreAddress = section.ShouldIgnoreAddress)
```

sarahelsaig (changes requested):

src/Modules/OrchardCore.Commerce.Payment/Services/CheckoutAddressService.cs
canadacubachina: HI @sarahelsaig : fix all after your review  except for total>10000 , i think 10000 is most credit card , stripe max payment allowed. maybe should put 10000 in the settings?

sarahelsaig: But 10000 of what? Do you mean 10000 USD? You are only checking the value, not the currency. In some currencies that's much less money. (e.g. 10000 HUF is only $28.74 ; 10000 JPY is $68.41 ; 10000 ISK is $81.31 ; 10000 CNY is $1392 ; All of these currencies are accepted by Stripe.)

We have to check if the total is zero (or negative, in case of misapplied discounts) because then we can skip payment processing entirely. We don't need to check for a maximum amount. If the total is higher than what the payment processor accepts, then they will reject the transaction which OCC can handle.

sarahelsaig (changes requested):

canadacubachina:
@sarahelsaig Please review it again, thanks

PR #609, opened 2025-06-27
Consider all front end clients

### LINKED ISSUE #608: Consider all front end situations
### Describe the bug

In .NET Blazor, when a user leaves a text input field blank, the value passed to the backend—such as an MVC controller—is typically an empty string, not null.

### To Reproduce

1. Go to 'PaymentService'
orderPart.BillingAddress.Address.Name

### Expected behavior

All front end clients

infofromca:
Fix #608


infofromca:
@sarahelsaig please review it

sarahelsaig (changes requested):

Issue #687, opened 2025-11-03
InvalidOperationException: Some, but not all products have gross price. This is invalid. (OCC-409)


infofromca:
## Describe the bug

A clear and concise description of what the problem is.

## To Reproduce

1. Edit 'Test Product'
2. Add  'Gross Price 10.00' and Tax Rate 20
3. do not add Tax to other product
4. See error, see pic

## Expected behavior

1--I know this is by code intentionally, but in real world, it is possible that one product (detergents) has tax , the other (for ex, vegetables) do not .
2-- More, it gave out broken page ( site)
## Screenshots

If app

<img width="2514" height="2019" alt="Image" src="https://github.com/user-attachments/assets/8797b35e-8d34-4fa3-87ef-f147a0fa6481" />

licable, add screenshots to help explain your problem.

[Jira issue](https://lombiq.atlassian.net/browse/OCC-409)

PR #692, opened 2025-11-04
Allow Tax Rate=0

### LINKED ISSUE #691: You must either provide both Gross Price and Tax Rate, or neither of them (OCC-413)
## Describe the bug

to resolve #687, there is a trick way, I can add gross price the same as PriceField, and tax rate as 0.
but now I got You must either provide both Gross Price and Tax Rate, or neither of them

## Expected behavior

can either provide  Gross Price or Tax Rate, or neither of them, or all of them

## Screenshots

[Jira issue](https://lombiq.atlassian.net/browse/OCC-413)

infofromca:
Fix #691

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -8,7 +8,7 @@
          should use Orchard Core references for the latest patch version to pull all versions up in the final app. -->
     <OrchardCoreVersion>2.1.0</OrchardCoreVersion>
 
-    <LombiqHelpfulLibrariesVersion>12.3.0</LombiqHelpfulLibrariesVersion>
+    <LombiqHelpfulLibrariesVersion>12.4.0</LombiqHelpfulLibrariesVersion>
     <LombiqTestsUIVersion>14.1.0</LombiqTestsUIVersion>
   </PropertyGroup>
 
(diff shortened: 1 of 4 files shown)


infofromca:
Blocker #687 :
we need to change the unit test logic (and its UI) here. 
because 
In Walmart of Canada, it sell computers with tax, but it also sell vegetables, milk, and other agri. products now on which there is no tax. and there are many many other physical and online stores.
after this change , I can apply this fix.

src/Modules/OrchardCore.Commerce/Handlers/TaxPartAndPricePartHandler.cs
```diff
@@ -39,7 +39,7 @@ public override async Task UpdatedAsync(UpdateContentContext context, PricePart
         var taxRate = taxPart.TaxRate?.Value ?? 0;
 
         var isGrossPricePresent = taxPart.GrossPrice?.Amount.IsValid == true;
-        var isTaxRatePresent = taxRate > 0;
+        var isTaxRatePresent = taxRate >= 0;
```
sarahelsaig: This is a bad change, because now `isTaxRatePresent` is always true, unless the admin manually enters a negative number. I've provided the fixed logic in #704.

sarahelsaig:
Thanks, your contribution has been fixed and merged via #704.

PR #729, opened 2026-04-12
NEST-608: Upgrade to Orchard Core 3.0 and .NET 10

sarahelsaig:
[NEST-608](https://lombiq.atlassian.net/browse/NEST-608)


arkadiuszwojcik:
Isn’t this ready to merge?

PR #730, opened 2026-04-29
OCC-221: Fix unnecessary payment intent creation and fix new obsoletes

### LINKED ISSUE #407: Multiple payment intents are created during a single session
### Describe the bug

Multiple payment intents are created in Stripe during a single session instead of reusing the already existing payment intents:

image

sarahelsaig:
[OCC-221](https://lombiq.atlassian.net/browse/OCC-221)
They are created in the background while you edit the addresses during checkout. It's because calls ~/checkout/price in case the prices need to be updated, which implicitly requests the payment provider data. The solution is to check if the price hasn't changed and the status is `requires_payment_method` (this is the status before you click on the payment provider button) then just use the stored payment intent instead.
Fixes #407


Skrypt:
Just a quick comment. I think  I needed to fix the fact that shopping cart cookies are not tenant aware. Maybe you will want to take a look if not already fixed.

sarahelsaig:
> Just a quick comment. I think I needed to fix the fact that shopping cart cookies are not tenant aware. Maybe you will want to take a look if not already fixed.

This PR didn't touch the shopping cart cookies so it's probably not fixed yet. Could you please open a new issue and elaborate on what kind of tenant awareness was necessary? If you can include a snippet or description of your fix that'd be great.

src/Modules/OrchardCore.Commerce.Payment.Stripe/Services/StripePaymentIntentService.cs
```diff
@@ -44,11 +48,18 @@ public async Task<PaymentIntent> GetPaymentIntentAsync(string paymentIntentId)
...
             await _requestOptionsService.SetIdempotencyKeyAsync(),
-            _hca.HttpContext.RequestAborted);
+            Aborted);
     }
 
     public async Task<PaymentIntent> CreatePaymentIntentAsync(Amount total, string shoppingCartId = null)
     {
+        var paymentIntentInfo = await _paymentIntentPersistence.RetrieveAsync(shoppingCartId);
+        if (paymentIntentInfo?.Amount == total &&
+            await GetPaymentIntentAsync(paymentIntentInfo.PaymentIntentId) is { Status: RequiresPaymentMethod } storedPaymentIntent)
```
wAsnk: Shouldn't we also check if the returned payment intent has the same amount? Here you only check if the cookie stored shopping cart has the same amount.

wAsnk (changes requested):

src/Modules/OrchardCore.Commerce.Payment.Stripe/Services/StripePaymentIntentService.cs
sarahelsaig: is _shouldn't_ be necessary unless something is weirdly corrupted.

wAsnk (approved):

PR #732, opened 2026-05-10
Avoids unnecessary allocations

### LINKED ISSUE #731: Follow the way of `LoginInfoJsonConverter` (OCC-439)
## Is your feature request related to a problem? Please describe

AmountConverter follow the way of LoginInfoJsonConverter of OC

## Describe the solution you'd like
Because OrchardCore uses a singleton instance of this converter when registering it into the JSON pipeline.
This pattern is intentional and has three benefits:
1. Avoids unnecessary allocations
 2. Ensures identity equality
✅ 3. Matches the pattern used by System.Text.Json itself
## Describe alternatives you've considered

[Jira issue](https://lombiq.atlassian.net/browse/OCC-439)

canadacubachina:
Fix #731

### DIFF
--- a/src/Libraries/OrchardCore.Commerce.MoneyDataType/Serialization/AmountConverter.cs
+++ b/src/Libraries/OrchardCore.Commerce.MoneyDataType/Serialization/AmountConverter.cs
@@ -10,6 +10,7 @@ namespace OrchardCore.Commerce.MoneyDataType.Serialization;
 
 public sealed class AmountConverter : JsonConverter<Amount>
 {
+    public static readonly AmountConverter Instance = new();
     public const string ValueName = "value";
     public const string CurrencyName = "currency";
     public const string Name = "name";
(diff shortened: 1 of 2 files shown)


sarahelsaig (changes requested):
In `LoginInfoJsonConverter` the instance is registered in code (see [here](https://github.com/OrchardCMS/OrchardCore/blob/94a6885328a25deea4202291650b43eb5e453d57/src/OrchardCore/OrchardCore.Users.Core/Extensions/UsersServiceCollectionExtensions.cs#L62)).
The created `Instance` variables here are not actually used.

infofromca:
> In `LoginInfoJsonConverter` the instance is registered in code (see [here](https://github.com/OrchardCMS/OrchardCore/blob/94a6885328a25deea4202291650b43eb5e453d57/src/OrchardCore/OrchardCore.Users.Core/Extensions/UsersServiceCollectionExtensions.cs#L62)). The created `Instance` variables here are not actually used.

As a lib or nuget, we can create it for the extensible for other custom module to extend by registration and DI

sarahelsaig:
> As a lib or nuget, we can create it for the extensible for other custom module to extend by registration and DI

That's not useful. You should never have to use these converters directly, because are tied to the converted class using an attribute: [`[JsonConverter(typeof(AmountConverter))]`](https://github.com/OrchardCMS/OrchardCore.Commerce/blob/d4725861b678199db3c769509d203362eb95115c/src/Libraries/OrchardCore.Commerce.MoneyDataType/Amount.cs#L16C1-L16C41).

If for some reason you still need an instance of the converter, you can create a static instance in your own code. Adding these static properties here would only be confusing for everyone else.

sarahelsaig:
Closing this and the issue for the above outlined reasons.

PR #744, opened 2026-07-14
Get Generator by different logic requirement with ISkuService

### LINKED ISSUE #742: generate sku logic (OCC-452)
## Describe the bug

  // If we have an SKU generator and the SKU is either empty or it must not be manually filled, then overwrite it
  // with the generated value.
  if (_skuGenerators.HighestPriority() is { } generator &&
      (string.IsNullOrWhiteSpace(part.Sku) || !generator.IsManualAllowed))
should be
  // If we have an SKU generator and the SKU is either empty or it must not be manually filled, then overwrite it
  // with the generated value.
  if (_skuGenerators.HighestPriority() is { } generator &&
      (string.IsNullOrWhiteSpace(part.Sku) && !generator.IsManualAllowed))

## To Reproduce

1. Go to 'line 49 of SkuValidationHandler'

## Expected behavior

Create a service with virtual method

## Screenshots

[Jira issue](https://lombiq.atlassian.net/browse/OCC-452)

### LINKED ISSUE #743: part.Sku = IsSkuReadOnly ? skuBefore : part.Sku.ToUpperInvariant(); not correct (OCC-453)
## Describe the bug
  // If the SKU is read-only then editing should not be possible, but here we undo any POST trickery just in case.
  part.Sku = IsSkuReadOnly ? skuBefore : part.Sku.ToUpperInvariant(); not correct

## To Reproduce

1. Go to 'line 57 of ProductPartDisplayDriver'

## Screenshots

[Jira issue](https://lombiq.atlassian.net/browse/OCC-453)

infofromca:
Fix #742 
Fix #743

make each method as virtual, so different logic can perform different implementation.


infofromca:
@sarahelsaig please review it

src/Modules/OrchardCore.Commerce/Services/SkuService.cs
```diff
@@ -0,0 +1,45 @@
...
+    {
+        if (!string.IsNullOrEmpty(skuBefore))
+        {
+            // If the SKU is read-only then editing should not be possible, but here we undo any POST trickery just in case.
+            part.Sku = IsReadOnly() ? skuBefore : part.Sku.ToUpperInvariant();
+        }
+        else
+        {
+            part.Sku = part.Sku.ToUpperInvariant();
+        }
```
sarahelsaig: If the read-only conditions are met, the method should simply reset the SKU and exit. This is to make it clear that no other logic should be checked if the SKU is read-only (in case additional checks or SKU processing may be added in the future).

```suggestion
        if (!string.IsNullOrEmpty(skuBefore) && IsReadOnly())
        {
            // If the SKU is read-only then editing should not be possible, but here we undo any POST trickery just in case.
            part.Sku = skuBefore;
            return;
        }

        part.Sku = part.Sku.ToUpperInvariant();
```

src/Modules/OrchardCore.Commerce/Services/SkuService.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            skuGenerator = null;
+            return false;
+        }
+
+        // Condition for allowing the generator
+        bool canGenerate = string.IsNullOrWhiteSpace(part.Sku) && !generator.IsManualAllowed;
+
+        skuGenerator = canGenerate ? generator : null;
+        return canGenerate;
```
sarahelsaig: When using the try get pattern and the result is false, the value of the `out` variable doesn't matter. You don't have to manually set it to `null`.

```suggestion
        skuGenerator = _skuGenerators.HighestPriority();

        // No generator is available.
        if (skuGenerator == null) return false;

        // Condition for allowing the generator.
        return string.IsNullOrWhiteSpace(part.Sku) && !skuGenerator.IsManualAllowed;
```

src/Modules/OrchardCore.Commerce/Services/SkuService.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        // No generator available
+        if (generator == null)
+        {
+            skuGenerator = null;
+            return false;
+        }
+
+        // Condition for allowing the generator
+        bool canGenerate = string.IsNullOrWhiteSpace(part.Sku) && !generator.IsManualAllowed;
```
sarahelsaig: This change is wrong. It should be with `||`, because
1. If the user didn't fill in the SKU even though manual entry is allowed, the system should generate one. (`true || false`)
2. If the user manually edited the HTML before submitting the form, but manual entry is not allowed, the system should overwrite the submitted value with a generated one (`false || true`).

Please revert the `&&` to `||`! Feel free to include the above explanation in a code comment, if you think it's unclear.

sarahelsaig (changes requested):

src/Modules/OrchardCore.Commerce/Services/SkuService.cs
```diff
@@ -0,0 +1,43 @@
...
+        if (!string.IsNullOrEmpty(skuBefore) && IsReadOnly())
+        {
+            part.Sku = skuBefore;
+            return;
+        }
+
+        part.Sku = part.Sku.ToUpperInvariant();
+    }
+
+    /// <summary>
```
sarahelsaig:
```suggestion
    /// <summary>Tries to retrieve an SKU generator, if possible.</summary>
    /// <returns>True when an <see cref="ISkuGenerator"/> should be used; otherwise, false.</returns>
    /// <remarks><para>
```

src/Modules/OrchardCore.Commerce/Services/SkuService.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// If the user didn't fill in the SKU even though manual entry is allowed, the system should generate one. (true || false).
+    /// If the user manually edited the HTML before submitting the form, but manual entry is not allowed,
+    /// the system should overwrite the submitted value with a generated one(false || true).
+    /// Keep virtual for different implementations to override the default behavior.
+    /// </summary>
+    /// <param name="part"><see cref="ProductPart"/>.</param>
+    /// <param name="skuGenerator">returns the <see cref="ISkuGenerator"/>.</param>
+    /// <returns>True when an <see cref="ISkuGenerator"/> should be used; otherwise, false.</returns>
```
sarahelsaig:
```suggestion
    /// </para></remarks>
```

infofromca:
@sarahelsaig please review it again, thanks

src/Modules/OrchardCore.Commerce/Services/SkuService.cs
sarahelsaig, line 1: You forgot to register this service in `Startup.ConfigureServices`.

sarahelsaig (changes requested):

